PR #5, opened 2018-11-01
Added Visual Studio Editor Script

ThePNDA:
Added the Script for creating a Visual Studio .json Highlighting File.
Also Changed up the gen_editor file to give a response on creating a file.
gen_editor now has 2 commands for creating the Visual Studio Script. [visualstudio, vs]

### DIFF
--- a/gen_editor
+++ b/gen_editor
@@ -1,18 +1,23 @@
 #! /usr/bin/env python3
 
 import sys
-from nml.editors import kate, notepadpp
+from nml.editors import kate, notepadpp, visualstudio
 
 def run():
-    if len(sys.argv) != 2 or sys.argv[1] not in ('kate', 'notepadpp'):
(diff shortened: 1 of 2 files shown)


planetmaker:
Nice :)
While I totally understand where the desire comes from to add some feedback to the gen_editor script, that is a behaviour change unrelated to the addition of VS editor file. It now behave more verbose than a "good" script usually is supposed to be (which means silent when everything works, but return a non-zero exit code, if things go wrong). Generelly, please don't mix-in unrelated additions in one PR :)

However I ignore this above advice and will merge this now. 

I would like to encourage you to make a subsequent PR which enables generation of the whole plug-in in the DevZone's build (basically the usability of your repo), so that users can just conveniently install it. I understand that VS requires more files, thus I'd suggest a change to make separate directories for each editor's output, e.g. editors/notepadpp, editors/visualstudio, editors/kate

PR #21, opened 2019-03-06
Eddi-nml branch for ActionC support

Eddi-z:


nml/actions/actionC.py
```diff
@@ -0,0 +1,33 @@
...
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA."""
+
+from nml.actions import base_action
+
+class ActionC(base_action.BaseAction):
+    def __init__(self, text):
+        self.text = text
+
+    def write(self, file):
+        #<Sprite-number> * <Length> 05 <type> <num-sprites> [<offset>]
```
PeterN: Wrong format for Action C

Eddi-z: must be a copy-pasta from somewhere

planetmaker:
Wouldn't it be more sensible to auto-convert any comment or any full-line-comment started with // in the nml source to a comment in the grf source? Possibly guarded by a cmd-line flag?

PR #24, opened 2019-03-16
Fix: Add missing parameter in industry and airport tiles in RandomAct…

stormcone:
…ion2.

If using a random_switch for an industry tile when the choice is based on its "related" object, an exception is thrown.

Test file is attached.
[test_industry.zip](https://github.com/OpenTTD/nml/files/2974138/test_industry.zip)


planetmaker:
Thanks for the fix!

PR #26, opened 2019-04-02
Fix: Allow access to towns (feature 0x12) as parent object

planetmaker:


frosch123:
What is this supposed to do?
Towns have no feature number.
NML and OTTD append a pseudo-feature at the end of all features for internal reasons, but there is no feature 0x12.

planetmaker:
> What is this supposed to do?
> Towns have no feature number.
> NML and OTTD append a pseudo-feature at the end of all features for internal reasons, but there is no feature 0x12.

There's a bug report in the forums: https://www.tt-forums.net/viewtopic.php?p=1220012#p1220012
That is you cannot write persistant town storage as parent from objects/industries. The easy fix seems to define an actual feature number for it....

PeterN:
0x12 should not be exposed as a usable feature, as far as I know it is virtual to work around some limitation.

planetmaker:
So, how do I access a town's persistant storage?

PeterN:
Where in the NewGRF specs?

planetmaker:
https://newgrf-specs.tt-wiki.net/wiki/VariationalAction2 uses consequently 0x12.

PeterN:
Thanks. That's rather one page, not "all through". And it's wrong.

frosch123:
I fixed the wiki.

PR #28, opened 2019-04-09
Fix: Allow access to towns as parent object

planetmaker:
So it needs some generic featureID used as parent.

Generally the features probably should be referenced via a symbolic name than their numeric ID in the code or this will fall onto our feet when a new feature is introduced. But that is a separate PR outside the scope of this.

And... I don't understand why this is not an update to the existing PR. Stupid github https://github.com/OpenTTD/nml/pull/26

### DIFF
--- a/nml/actions/action0properties.py
+++ b/nml/actions/action0properties.py
@@ -16,6 +16,7 @@ with NML; if not, write to the Free Software Foundation, Inc.,
 import itertools
 from nml import generic, nmlop
 from nml.expression import BinOp, ConstantNumeric, ConstantFloat, Array, StringLiteral, Identifier, ProduceCargo, AcceptCargo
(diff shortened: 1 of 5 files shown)


PeterN:
I think that separate PR should happen before this, to be honest.

frosch123:
Alternative solution:
https://gist.github.com/frosch123/fb62db8e6a65c0de1614ad1f663f1827

It uses a text instead of a (wrong) number.
Towns are accepted by feature_name, but not by parse_feature.
Removes the references to feature 8, which were likely means to be "towns".

glx22:
Since checks are enabled, a rebase to master is required to include workflow file.

LordAro:
No change in a while. Gonna close this

PR #37, opened 2019-05-04
Fix #36: global variable traffic_side was inconsistent with global pa…

### LINKED ISSUE #36: Wrong variable 'traffic_side'
VarAction2 variable 0x06 is 0x10 when traffic side is right, not 0x01.

Eddi-z:
…rameter traffic_side


Eddi-z:
The global parameter is defined with:
`./nml/global_constants.py:    'traffic_side'                       : {'param': 0x86, 'bit': 4},`
for comparison with the constants
```
./nml/global_constants.py:    'TRAFFIC_SIDE_LEFT'                  : 0,
./nml/global_constants.py:    'TRAFFIC_SIDE_RIGHT'                 : 1,
```

This PR changes the definition in `./nml/actions/action2var_variables.py` to match

PR #53, opened 2019-10-19
Fix #52: time.clock() has been removed in python 3.8

### LINKED ISSUE #52: nmlc doesn't work with python 3.8
time.clock has been removed in python 3.8, generating `Error:    (AttributeError) "module 'time' has no attribute 'clock'".`

glx22:


nielsmh:
This is used to update progress messages once in a while, right? Wouldn't a wall clock time instead of a process time stopwatch be better? E.g. `time.monotinic()`.

PR #54, opened 2019-10-23
Fix #39: Add compatibility with >=pillow-7.0.0

### LINKED ISSUE #39: (AttributeError) "module 'PIL.Image' has no attribute 'VERSION'".
I am working on a Windows 64 bit laptop with Conda v2019.03, running a Python 3.6.0 environment. After installing all the dependencies, when I try to run `nmlc -h` it shows me the following error:
`nmlc ERROR: nmlc: An internal error has occurred:
nmlc-version: 0.4.5
Error:    (AttributeError) "module 'PIL.Image' has no attribute 'VERSION'".
Command:  ['C:\\ProgramData\\Anaconda3\\envs\\py3.6\\Scripts\\nmlc']
Location: File "c:\programdata\anaconda3\envs\py3.6\lib\site-packages\nml\version_info.py", line 74, in get_lib_versions`

**Workaround**
After a quick round of googling, I found that the issue is with Pillow v6.0.0. Uninstalling it and doing `pip install pillow==5.4.1` solved the issue for me.

I hope this helps anyone else who might have the same issue.

ConiKost:
Changed PILLOW_VERSION to PIL.__version__,
as it got removed with >=pillow-7.0.0.

In order not to break compatibility with older versions,
PILLOW_VERSION will be used as a fallback.

Closes: #39
Signed-off-by: Conrad Kostecki <conrad@kostecki.com>

### DIFF
--- a/nml/version_info.py
+++ b/nml/version_info.py
@@ -139,11 +139,17 @@ def get_lib_versions():
     #PIL
     try:
         from PIL import Image
-        versions["PIL"] = Image.PILLOW_VERSION
+        try:
+            versions["PIL"] = Image.__version__
+        except AttributeError:
+            versions["PIL"] = Image.PILLOW_VERSION
     except ImportError:
         try:
             import Image
-            versions["PIL"] = Image.PILLOW_VERSION
+            try:
+                versions["PIL"] = Image.__version__
+            except AttributeError:
+                versions["PIL"] = Image.PILLOW_VERSION
         except ImportError:
             versions["PIL"] = "Not found!"
 


Eddi-z:
Commit message should probably start with "Fix #39: ..."

Also, does it still work with older pillow, or is this a new minimum requirement?

ConiKost:
> Also, does it still work with older pillow, or is this a new minimum requirement?

At least, with version 6.x is working. I can't test version 5 and older.

nml/version_info.py
```diff
@@ -139,11 +139,11 @@ def get_lib_versions():
     #PIL
     try:
         from PIL import Image
-        versions["PIL"] = Image.PILLOW_VERSION
+        versions["PIL"] = Image.__version__
```
Eddi-z: could do
```
try:
  old version
except AttributeError:
  new version
```

nml/version_info.py
```diff
@@ -139,11 +139,11 @@ def get_lib_versions():
...
     except ImportError:
         try:
             import Image
-            versions["PIL"] = Image.PILLOW_VERSION
+            versions["PIL"] = Image.__version__
```
Eddi-z: Also, maybe, while here, try to remove this code duplication by restructuring something?

ConiKost:
Thanks. But ```PILLOW_VERSION```, as in current master, works for you?

planetmaker:
> Thanks. But `PILLOW_VERSION`, as in current master, works for you?

Yes: 
1& [ingo:~/ottd/grfdev/nml/nml] 78cd84c+* ± nmlc --version
2019-10-20-g78cd84ce
Library versions encountered:
PLY: 3.9
PIL: 4.0.0
1& [ingo:~/ottd/grfdev/nml/nml] 78cd84c+* ± ppython3 -V
Python 3.5.3

nml/version_info.py
ConiKost: Allright, added. Seems to work here.

nml/version_info.py
ConiKost: TBH, I have no clue, how to write this better, since it checks for 2 different imports. Sorry, I am not a native python coder.

planetmaker:
> @planetmaker Could you try latest variant?

It's PILLOW_VERSION (you are missing one L). That fixed it works for me.

ConiKost:
> It's PILLOW_VERSION (you are missing one L). That fixed it works for me.

already fixed ;-)

matthijskooijman:
> That patch does not fly as it breaks nml for older versions of PIL / pillow:
> PIL: 1.1.7

But the README says the minimum supported pillow version is 5.2, so why is this relevant? It will (I expect) then just break elsewhere? Or are older versions actually supported after all?

ConiKost:
PIL =! Pillow. Version ```1.1.7``` is for historic reasons.

Pillow 5.2.0 also reports ```1.1.7``` with ```PIL.VERSION```.
-> https://pillow.readthedocs.io/en/stable/releasenotes/5.2.0.html

matthijskooijman:
> PIL =! Pillow. Version 1.1.7 is for historic reasons.

Ah, of course, I misread the comment then. @planetmaker, what pillow version were you using when you posted:

>  Error: (AttributeError) "module 'PIL.Image' has no attribute 'version'".


Looking more closely, it seems this PR now uses `PIL.Image.__version` (which is apparently not available in the version the @planetmaker tested). However, IIUC, the docs recommend using `PIL.__version__` instead, which should be available since version 3.4 (see also https://github.com/OpenTTD/nml/issues/39#issuecomment-544713669)

matthijskooijman:
@LordAro I'm not actually sure that the code is now checking the right (or at least recommended) version number, see my previous comment. @ConiKost, what do you think?

ConiKost:
@matthijskooijman I guess, it depends on the specific version, which he has.

But currently, as far I understand the code, both, `PIL.Image.__version__` and `PIL.__version__` are tried, before doing a fallback. So this should be fine, as it is?

matthijskooijman:
AFAICS, `PIL.__version__` is never tried, only `PIL.Image.__version__` and `Image.__version__`.

https://github.com/OpenTTD/nml/blob/abb7a3a71cf83f39d124edf2e6526678e53b1c38/nml/version_info.py#L140-L154

Looking at the code gain, I also wonder if the fallback to `import Image` is still needed? I haven't looked in detail, but before I found that `PIL.__version__` was introduced in pillow 3.4.0 (https://github.com/python-pillow/Pillow/commit/7bfb2527bb0b832f830f87c37bd165028118e442) and it seems that `PIL.Image` was also available already in that version (and the minimium pillow version for nml is now 5.something, right?).

ConiKost:
FWICS, you seem to be right, that `import Image` could not be needed anymore.
About version 5, I wasn't the one, who said said ;-) I only was asked, to add 5.2 as a minimum version, so I did. I don't know the code, so I can't judge on that.

matthijskooijman:
Thanks. I've opened PR #59 further simplifying the Pillow imports. As for the minimum version, I've opened issue #58 for that.

PR #55, opened 2019-11-07
Codechange: use pyinstaller to generate standalone executable (and other mostly windows related changes)

glx22:
cx_Freeze was not working correctly, nml_lz77 lib was built and included but not used.
pyinstaller includes and uses the lib, as long as it's next to `nmlc`, this is enforced in `nmlc.spec`.
`python ./setup.py install` was not really usable on windows, as it copied `nmlc` directly, now setuptools will generate a wrapper executable, as it did before 78cd84c.
The batch file just builds the lib at the right place then creates the standalone exe in `dist`.

### DIFF
--- a/build-dist.bat
+++ b/build-dist.bat
@@ -1,0 +1,2 @@
+python .\setup.py build_ext --inplace
+pyinstaller .\nmlc.spec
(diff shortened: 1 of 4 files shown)


glx22:
Some generated executables, using different python version, are available as artifacts [here](https://github.com/glx22/nml/commit/a3989132de7a276e7096d32ef9b15cb0a9462c32/checks?check_suite_id=278199092)

glx22:
Hmm just thinking, but maybe it would be nice to set `install_requires` in setup.py.

PR #62, opened 2019-11-20
Mechanical cleanups

FLHerne:
A couple of trivial search-and-replace cleanups that should have no functional changes.

Tests pass.


andythenorth:
Tested locally:

- builds
- tests pass
- sample newgrf compile works
- diff looks fine, all looks like legacy artefacts

PR #63, opened 2019-11-21
Allow PLY to generate parsing/lexing tables.

FLHerne:
...and modify setup.py to create them at build time.

On my laptop, pre-generated tables reduce the time spent in
 `NMLParser.__init__()` from 2.2s to 0.2s, halving the total runtime
 of nmlc for small input files.

PLY will attempt to regenerate these automatically at runtime if the
 grammar has changed. In the case where they're system-installed and not
 writeable, the grammar is also system-installed and not writeable. :P


glx22:
A small change is required for pyinstaller. I made the changes locally and can push them to your branch if it's ok.

FLHerne:
@glx22 Yes, please go ahead. :-)

andythenorth:
I tested this with Iron Horse and  --verbosity=4 flag.  Python is 3.5.

Results:
- Init parser: consistently 0.0s with this PR, 0.4-0.5s in master
- Parsing: less consistent results, likely due to other system load.  With this PR 16.4-17.3s.  Without this PR 17.0-20.0s.

glx22:
Hmm `pip install -e .` or `python setup.py develop` fails with the setup.py changes if ply is not yet installed. Without the changes the dependencies are installed and everything is fine.

setup.py
```diff
@@ -3,9 +3,18 @@
 import os
 
 from setuptools import setup, Extension, find_packages
-from nml import version_info
+from setuptools.command.build_py import build_py
+from nml import parser, version_info
```
glx22:
```suggestion
from nml import version_info
```

setup.py
```diff
@@ -3,9 +3,18 @@
...
 NML_VERSION = version_info.get_and_write_version()
 
+
+class NMLBuildPy(build_py):
+    def run(self):
+        # Create a parser so that nml/generated/{parse,lex}tab.py are generated.
```
glx22:
```suggestion
        from nml import parser
        # Create a parser so that nml/generated/{parse,lex}tab.py are generated.
```

FLHerne: Thanks. This fix was what I thought of when I saw your last comment, but I hadn't got to it yet. :P

PR #64, opened 2019-11-22
Fix: enable VT100 sequences on windows

glx22:
`[K` on my screen were annoying so instead of removing them I enabled support for them. As VT100 sequences are now enabled, color codes can be used too so I re-enabled the warning color and I changed it to bright yellow because it seems normal yellow is the default powershell foreground (even if visually it's some pale white). And I added color to error messages (bright red, because red is hard to read on default blue powershell background)

Commits should probably be squashed, but that can happen when merging.

### DIFF
--- a/nml/generic.py
+++ b/nml/generic.py
@@ -16,6 +16,27 @@ with NML; if not, write to the Free Software Foundation, Inc.,
 # -*- coding: utf-8 -*-
 import sys, os, time
 
+# Enable VT100 sequences on windows consoles
+if os.name == "nt":
+    class Break(Exception): pass
+    try:
+        from ctypes import byref, windll
+        from ctypes.wintypes import DWORD, HANDLE
+        kernel32 = windll.kernel32
+        h = kernel32.GetStdHandle(-11) # stdout
+        if h is None or h == HANDLE(-1):
+            raise Break()
+        FILE_TYPE_CHAR = 0x0002
+        if (kernel32.GetFileType(h) & 3) != FILE_TYPE_CHAR:
+            raise Break()
+        mode = DWORD()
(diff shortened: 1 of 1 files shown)


nml/generic.py
```diff
@@ -16,6 +16,26 @@
...
 
+# Enable VT100 sequences on windows consoles
+if os.name == "nt":
+    try:
+        from ctypes import byref, windll
+        from ctypes.wintypes import DWORD, HANDLE
+        kernel32 = windll.kernel32
+        h = kernel32.GetStdHandle(-11) # stdout
+        if h is None or h == HANDLE(-1):
+            raise
```
LordAro: is a bare "raise" when not inside an except block valid python?

glx22: hmm "If no expressions are present, raise re-raises the last exception that was active in the current scope. If no exception is active in the current scope, a RuntimeError exception is raised indicating that this is an error.", so it does work for what I want, ie early exit

LordAro: Fair enough, bit unusual though... (And bare `except` is frowned upon as well - maybe `except RuntimeError` ? )

glx22: another option is
```
while True:
    try:
        ...
        if exit_early:
            break
        ...
        break
    finally:
        pass
```

glx22: hmm no need to try..finally in this case

nml/generic.py
```diff
@@ -16,6 +16,24 @@
 # -*- coding: utf-8 -*-
 import sys, os, time
 
+# Enable VT100 sequences on windows consoles
+while os.name == "nt":
```
LordAro: No, I think this is worse, makes things much less clear. Is a few extra levels of indentation the worst thing in the world?

PR #65, opened 2019-11-23
Fix #57: Check coherency of GRF parameters limits

### LINKED ISSUE #57: GRF integer parameters can be given negative values.
When giving a grf parameter a negative value, the compiler doesn't complain. Though OpenTTD has an assertion check to see if the values of the dropdown are positive.

Example:
```
min_value: -1;
max_value: 1;
def_value: 0;
```

Image of Yaktocat

A zip containing the .grf and the .nml file.
[grf.zip](https://github.com/OpenTTD/nml/files/3854384/grf.zip)

Not sure, but maybe this is a OpenTTD related issue. This issue exists in various versions of OpenTTD, aswell as in JGR.

glx22:
Added a `uvalue` attribute to `ConstantNumeric` containing unsigned equivalent of value. Using this attribute it's easy to check coherency of min/max. Also used `uvalue` to ensure the parameter settings are in the uint32 range.

### DIFF
--- a/nml/actions/action14.py
+++ b/nml/actions/action14.py
@@ -263,10 +263,10 @@ def param_desc_actions(root, params):
             else:
                 assert setting.type == 'bool'
                 setting_node.subnodes.append(BinaryNode("TYPE", 1, 1))
-                bit = setting.bit_num.uvalue if setting.bit_num is not None else 0
+                bit = setting.bit_num.value if setting.bit_num is not None else 0
                 setting_node.subnodes.append(SettingMaskNode(param_num, bit, 1))
             if setting.def_val is not None:
-                setting_node.subnodes.append(BinaryNode("DFLT", 4, setting.def_val.uvalue))
+                setting_node.subnodes.append(BinaryNode("DFLT", 4, setting.def_val.value))
             param_root.subnodes.append(setting_node)
             setting_num += 1
         param_num += 1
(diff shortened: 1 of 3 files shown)


nml/actions/action14.py
```diff
@@ -258,10 +262,10 @@ def param_desc_actions(root, params):
             else:
                 assert setting.type == 'bool'
                 setting_node.subnodes.append(BinaryNode("TYPE", 1, 1))
-                bit = setting.bit_num.value if setting.bit_num is not None else 0
+                bit = setting.bit_num.uvalue if setting.bit_num is not None else 0
```
glx22: probably useless change

nml/actions/action14.py
```diff
@@ -258,10 +262,10 @@ def param_desc_actions(root, params):
...
                 setting_node.subnodes.append(SettingMaskNode(param_num, bit, 1))
             if setting.def_val is not None:
-                setting_node.subnodes.append(BinaryNode("DFLT", 4, setting.def_val.value))
+                setting_node.subnodes.append(BinaryNode("DFLT", 4, setting.def_val.uvalue))
```
glx22: probably useless change

nml/ast/grf.py
```diff
@@ -241,7 +241,7 @@ def set_property(self, name, value):
         self.properties_set.add(name)
         if name == 'names':
             for name_value in value.values:
-                num = name_value.name.reduce_constant().value
+                num = name_value.name.reduce_constant().uvalue
```
glx22: probably useless change

glx22:
hmm I think I should also check that default value is inside min/max

PR #66, opened 2019-11-24
Add: allow use of switches and random switches as procedures

glx22:
NFO output seems correct, but ID handling and generation needs work. For now 2 different procedures can have the same ID.
Work probably needed in other areas too. That's why it's a draft.

Feel free to add commits in my branch.


andythenorth:
I've tested this with a single stored procedure in Iron Horse.
Branch is https://github.com/andythenorth/iron-horse/tree/stored-procedure-test-haulage-speed-bonus

This replaces global-switch-and-registers with checking a procedure result.  It makes more readable code.  It's not a very interesting test case but eh.  Appears to work.

glx22:
I think it's ready now.

nml/expression/array.py
```diff
@@ -31,3 +31,9 @@ def __str__(self):
 
     def reduce(self, id_dicts = [], unknown_id_fatal = True):
         return Array([val.reduce(id_dicts, unknown_id_fatal) for val in self.values], self.pos)
+
+    def collect_references(self):
+        refs = []
+        for v in self.values:
+            refs += v.collect_references()
+        return refs
```
LordAro:
```suggestion
        return [v.collect_references() for v in self.values]
```
would be slightly more pythonic

nml/expression/bitmask.py
```diff
@@ -39,5 +39,11 @@ def reduce(self, id_dicts = [], unknown_id_fatal = True):
             ret = BinOp(nmlop.OR, ret, val, self.pos)
         return ret.reduce()
 
+    def collect_references(self):
+        refs = []
+        for v in self.values:
+            refs += v.collect_references()
+        return refs
```
LordAro: similarly here

nml/expression/array.py
ldpl: more like sum((v.collect_references() for v in self.values), [])
or itertools.chain.from_iterable for max effeciency

andythenorth:
One final +1: I have converted Iron Horse to use procedures widely, and have compiled hundreds of times now using nmlc built with this PR.  Zero issues so far.

PR #68, opened 2019-12-01
Fix: close image files after use during palette check

andythenorth:
With pypy, nmlc trivially trips file handle limits on macOS during palette checks, resulting in: "[Errno 24] Too many open files".

This is a common enough problem in python, which can be resolved by either

- end user setting `ulimit -n 4096` on a per shell basis
- closing PIL Image files with `Image.close()`, where `Image.open()` has been used 

_This PR closes images opened for palette checks, which is a trivial case.  Errno 24 no longer triggers for me. Tests pass._

Errno 24 does not trigger for me with any cPython 3.x, but pypy cuts run times by up to 50% in my testing compared to cPython 3.8, so we do want to be compatible with pypy.

There are a couple of other uses of Image.open() but they looked less trivial, and they're not tripping Errno 24 for me.

For the record, `Image.load()` self-closes so wouldn't have the same issue, but `Image.open()` is often the wanted command, especially for reading metadata.

### DIFF
--- a/nml/main.py
+++ b/nml/main.py
@@ -335,6 +335,7 @@ def nml(inputfile, input_filename, output_debug, outputfiles, start_sprite_num, 
         if im.mode != "P":
             continue
         pal = palette.validate_palette(im, f)
+        im.close()
 
         if forced_palette != "ANY" and pal != forced_palette and not (forced_palette == "DEFAULT" and pal == "LEGACY"):
             raise generic.ImageError("Image has '{}' palette, but you forced the '{}' palette".format(pal, used_palette), f)


nielsmh:
I'm wondering if using a `with` block (context manager) isn't the more correct thing here? Assuming the Image class supports it, but I'd expect it to.

andythenorth:
See also https://pillow.readthedocs.io/en/latest/reference/open_files.html#file-handling

planetmaker:
I agree with fsj: the closing should be done by means of using the with contextmanager
In good script kiddie manner, here's some examples:
https://stackoverflow.com/questions/713794/catching-an-exception-while-using-a-python-with-statement

andythenorth:
Thanks for reviews.   Bearing in mind that I am a clown-shoes python programmer who does not know how to use a contextmanager, current options are:

* I learn how to implement a contextmanager in combination with try / except, which is maybe trivial, but I do not know how to do it (the 'try/except' makes it not a trivial use of 'with')
* someone else adds the context manager
* we ship my solution, and leave a note that there are better ways
* we leave nmlc 'broken' on macOS with larger grfs, with a viable workaround of 'ulimit -n 4096'

LordAro:
Superseded.

PR #70, opened 2019-12-07
Reduce boilerplate for creating BinOps

FLHerne:
Should be no functional changes, except a few line positions might be tracked implicitly.

Last patch created mostly with some horrible sed expressions.

Regression tests pass.

### DIFF
--- a/nml/expression/binop.py
+++ b/nml/expression/binop.py
@@ -90,7 +90,7 @@ class BinOp(Expression):
         # - If the operator allows it and the second expression is more complex than
         #   the first one swap them.
         op = self.op
-        if op in commutative_operators or self.op in (nmlop.CMP_LT, nmlop.CMP_GT):
+        if op.commutative or op in (nmlop.CMP_LT, nmlop.CMP_GT):
             prio1 = self.get_priority(expr1)
             prio2 = self.get_priority(expr2)
             if prio2 < prio1:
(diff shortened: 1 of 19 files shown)


nml/expression/binop.py
```diff
@@ -90,7 +90,7 @@ def reduce(self, id_dicts = [], unknown_id_fatal = True):
         # - If the operator allows it and the second expression is more complex than
         #   the first one swap them.
         op = self.op
-        if op in commutative_operators or self.op in (nmlop.CMP_LT, nmlop.CMP_GT):
+        if op.commutative or op in (nmlop.CMP_LT, nmlop.CMP_GT):
```
planetmaker: Would it make sense to define a set (or another property like communative) called 'comparison' or similar?

FLHerne:
I'd like to make some changes to this when I get time, please don't merge just yet. ;-)

nml/expression/binop.py
FLHerne: Just a "comparison" flag isn't really helpful, because you'd still need a table of which one to replace it with.

I looked at adding a `swapped_operand_op` property to cover this, but it's awkward because Python doesn't have forward references...

There are several things in this file that could be done more nicely, but I don't think it's particularly in scope for this patch.

FLHerne:
Rebased. I decided the changes I wanted to make were a mistake, so this is pretty much the same as before. I've tested with more grfs that this doesn't change the output.

If possible I'd like to get this in before #103, because the rebase would be a nightmare. If that means I owe @LordAro to help rebase that instead, so be it, I think it would be easier...

Yexo:
Nice cleanup, the new code is much easier to follow without all the boilerplate.

FLHerne:
Note: I did consider implementing actual operators on `Expression`, but ultimately decided it would be confusing.

glx22 (approved):
Adds a new CodeQL error, but this can be handled later.
Code is way easier to read (and write).

FLHerne:
Thanks.

I'm confident the circular import isn't an actual problem -- nothing from `binop` is actually used in execution within `nmlop` or vice versa at import time. Moving the import inside the function would be a performance hit, and all my other ideas are intrusive enough to be another PR.

PR #71, opened 2019-12-07
Add: nmlc option to force regeneration of parser tables

glx22:
Since merge of #63, parser tables are not regenerated when parser and lexer are modified. This new option works around it without disabling optimisations.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ class NMLBuildPy(build_py):
     def run(self):
         # Create a parser so that nml/generated/{parse,lex}tab.py are generated.
         from nml import parser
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -12,7 +12,7 @@ class NMLBuildPy(build_py):
     def run(self):
         # Create a parser so that nml/generated/{parse,lex}tab.py are generated.
         from nml import parser
-        parser.NMLParser()
+        parser.NMLParser(True)
```
FLHerne: `parser.NMLParser(rebuild=True)` for clarity?

PR #35, opened 2019-05-01
Fix ottd_display_speed to reflect changes done in OpenTTD

juzza1:
https://github.com/OpenTTD/nml/commit/736f034b4d4ec9e9548247ad0126f08d1db0af7c fixed rounding errors with nml->nfo conversion, however the [ottd_display_speed](https://github.com/OpenTTD/nml/blob/f146d89c5366ec1bbc8f9351ef40b3a59c115024/nml/actions/action0properties.py#L257) function has not been updated to reflect changes done in https://github.com/OpenTTD/OpenTTD/commit/76344163c7d190894629768549130438c34174aa. Specifically, the display value should be divided only after all other conversions have been done. The updated version of the function fixes the many off-by-one rounding errors currently present with nml coded vehicles.

PR #75, opened 2019-12-20
Fix ottd_display_speed to reflect changes done in OpenTTD

planetmaker:
This is a rebased version of https://github.com/OpenTTD/nml/pull/35

Closes #35


nml/actions/action0properties.py
```diff
@@ -255,7 +255,7 @@ def cargo_list(value, max_num_cargos):
 }
 
 def ottd_display_speed(value, divisor, unit):
-    return int(value.value / divisor) * 10 // 16 * unit.ottd_mul >> unit.ottd_shift
+    return (round(value.value // divisor * 10 * unit.ottd_mul) >> unit.ottd_shift) // 16
```
frosch123: Why is there an integer division "// divisor" inside a round? That looks wrong to me.

frosch123: I think this function is meant to replicate the exact behaviour of OTTD's ConvertKmhishSpeedToDisplaySpeed.
Thus I would expect:
`(round(value.value / divisor * 10 * unit.ottd_mul) >> unit.ottd_shift) // 16`

planetmaker: I was actually wondering whether we may have a round there at all... or whether we get some platform dependency there which we might not want.
Though without an integer being left of the shift operator, we get a type error...

PR #82, opened 2020-02-16
Feature: global var 0x21/0xA1: openttd_version, openttd_major, openttd_minor, openttd_revision, openttd_build

glx22:


frosch123:
openttd_version and helper-function version_openttd already exist:
https://newgrf-specs.tt-wiki.net/wiki/NML:General#General_variables

What's the use-case for adding single variables?

glx22:
Oh then https://wiki.openttd.org/NewGRF_Specification_Status is wrong

PR #85, opened 2020-02-17
Change: industry tiles and houses now support 16 accepted cargos

glx22:


andythenorth:
This PR adds the last thing needed for newgrf support of OpenTTD 1.9.0 :) 

https://wiki.openttd.org/NewGRF_Specification_Status

nml/actions/action0properties.py
```diff
@@ -291,6 +295,26 @@ def ctt_list(prop_num, *values):
...
+    cargos = []
+    for value in values:
+        if not isinstance(value, Array) or len(value.values) > 16:
+            raise generic.ScriptError("accepted_cargos must be an array with no more than 16 values", value.pos)
+        tile_cargos = []
+        for cargo_amount_pair in value.values:
+            if not isinstance(cargo_amount_pair, Array) or len(cargo_amount_pair.values) != 2:
+                raise generic.ScriptError("Each element of accepted_cargos must be an array with two elements: cargoid and amount", cargo_amount_pair.pos)
+            cargo_id = cargo_amount_pair.values[0].reduce_constant().value
+            cargo_amount = cargo_amount_pair.values[1].reduce_constant().value
```
frosch123: How about a 
```
 if isinstance(cargo_amount, ConstantNumeric):
        generic.check_range(cargo_amount.value, 0, 8, "cargo acceptance", cargo_amount.pos)
```
cargo_id could also be checked for range, but probably does not help much.

glx22: for industry tiles it can be <0 to counter `INDTILE_FLAG_ACCEPT_ALL`  effect

frosch123: good point, I forgot that :)

PR #91, opened 2020-02-24
Fix #90, e9429c89: Generated procedure calls where not properly marked

### LINKED ISSUE #90: Length callback for vehicles doesn't accept switches
So I've built the master branch of NML and now it seems that all vehicles do not accept switches for the `length` Action3 callback.
My source for the code can be found here: https://github.com/spnda/VGF-Set
(I use a python script to put all pnml files together, for the file in question look at [src/trams/Type-P.pnml](https://github.com/spnda/VGF-Set/blob/master/src/trams/Type-P.pnml).

If `length` is given a switch, nmlc throws an error. However there's no error thrown when there is just a value passed.
```
length: 5;
```

The error I receive:
```
 nmlc ERROR: "src/trams/Type-P.pnml", line 60: Unexpected identifier encountered: 'switch_type_p_length'
Included from: "vgf.pnml", line 1
Included from: "src/trams/Type-O.pnml", line 59
Included from: "vgf.pnml", line 1
Included from: "src/trams/Type-KEE.pnml", line 49
Included from: "vgf.pnml", line 2
Included from: "src/template.pnml", line 29
Included from: "vgf.pnml", line 1
Included from: "src/sort.pnml", line 15
Included from: "vgf.pnml", line 1
Included from: "src/header.pnml", line 9
Included from: "vgf.pnml", line 1
Included from: "src/checks.pnml", line 8
Included from: "vgf.pnml", line 1
Included from: "src/cargotable.pnml", line 6
Included from: "vgf.pnml", line 1
Included from: "src/basecost.pnml", line 8
Included from: "vgf.pnml", line 1
```

glx22:
When I added the check I totally missed it was possible to create procedure calls when generating action 3


glx22:
Good idea

PR #93, opened 2020-04-02
Add release workflow

LordAro:
Not finished yet

* [x] Build Python wheels for Windows, MacOS & Linux (manylinux2010)
* [x] Builds standalone (win64) nmlc executable
* [x] Fixes a bug in git version calculation with tags
* [x] Make py38 work ( RalfG/python-wheels-manylinux-build#14 )
* [x] Work out which artefacts want to be distributed with the release (fixes Linux release failure)
* [x] ~See if build artefacts can be attached to the release~ Not worth the bother
* [x] Remove test code
* [x] Add `PYPI_PASSWORD` secrets to repo

Latest release build example: https://github.com/LordAro/nml/actions/runs/83157723

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -52,7 +52,6 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-        architecture: ${{ startsWith(matrix.os, 'windows') && 'x86' || 'x64' }}
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,101 @@
+name: Release
+
+on:
+  push:
+    branches:
+    - master
+    tags:
+    - '*'
+
+jobs:
+  release:
+    name: Python 3.x on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-2016]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/cache@v1
+      if: startsWith(matrix.os, 'ubuntu')
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+
+    - uses: actions/cache@v1
+      if: startsWith(matrix.os, 'macOS')
+      with:
+        path: ~/Library/Caches/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+
+    - uses: actions/cache@v1
+      if: startsWith(matrix.os, 'windows')
+      with:
+        path: ~\AppData\Local\pip\Cache
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+        pip install -e .
+
+    - name: Build manylinux Python wheels
+      if: startsWith(matrix.os, 'ubuntu')
+      uses: RalfG/python-wheels-manylinux-build@v0.2.2-manylinux2014_x86_64
+      with:
+        python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38'
+        build-requirements: '-e .'  # pip args
+
+    - name: Publish manylinux Python wheels
+      if: startsWith(matrix.os, 'ubuntu')
+      env:
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        twine upload --username __token__ wheelhouse/*-manylinux*.whl
+
+    - name: Build and publish wheel
+      if: "! startsWith(matrix.os, 'ubuntu')"
+      env:
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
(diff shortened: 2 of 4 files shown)


.github/workflows/regression.yml
```diff
@@ -64,3 +64,10 @@ jobs:
       shell: bash
       env:
         NMLC: nmlc
+
+    # Just check we can successfully build it
```
TrueBrain, line 68: for other projects, we do this in "testing.yml", as there might be other things to test for (like flake, black, ..)

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
+name: Release
+
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  regression:
+
```
TrueBrain, line 10: silly newline

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
+name: Release
+
+
```
TrueBrain, line 3: silly newline

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+on:
+  release:
+    types: [created]
```
TrueBrain, line 6: In other projects we do this on "push/tags". Not sure if they differ. Might be good to either change them all to the one or the other.

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+        twine upload --repository-url https://test.pypi.org/legacy/ dist/*.whl
+
+    - name: Build standalone executable
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        pip install --upgrade pyinstaller
+        pyinstaller ./nmlc.spec
+        $version = python ./setup.py --version
+        echo "::set-output name=version::$version"
+      id: exe
```
TrueBrain, line 87: I am sure this id can be better.

nml/version_info.py
```diff
@@ -119,8 +119,8 @@ def get_git_version(detailed = False):
         # Compose the actual version string
         str_tag = ""
         if len(tag) > 0:
```
TrueBrain, line 121: while at it, fix this line too: `if tag:` :)

setup.py
```diff
@@ -17,7 +17,7 @@ def run(self):
         super().run()
 
 setup(
-    name='nml',
+    name='nml-test',
```
TrueBrain, line 20: temporary, I assume?

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+      with:
+        path: ~\AppData\Local\pip\Cache
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
```
glx22, line 48: Regression uses `architecture: ${{ startsWith(matrix.os, 'windows') && 'x86' || 'x64' }}`, release defaults to `architecture: 'x64'` as it's not specified.

planetmaker: it needs python >= 3.5 at least

.github/workflows/regression.yml
LordAro: out of scope for this PR, I feel :p

.github/workflows/release.yml
LordAro: "3.x" is just "the latest", afaik

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
+name: Release
+
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  regression:
```
TrueBrain, line 9: too much copy/pasting :D

setup.py
LordAro: From the PR description
> * [ ] Remove test code

Yes :p Necessary for testing upload to test.pypi.org, as nml already exists (surprisingly enough)

.github/workflows/release.yml
TrueBrain: The only thing I prefer is that we do the same thing for all projects. So pick one and apply it everywhere. This is not about personal preferences, this is about keeping things sane :)

MinchinWeb:
Oooh... does this mean we'll start seeing "regular" PyPI releases?

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+on:
+  push:
+    branches:
+    - master
+    tags:
+    - '*'
+  repository_dispatch:
+    types:
+    - publish_latest_tag
+    - publish_master
```
TrueBrain, line 12: are you using the dispatch?

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+        python setup.py sdist bdist_wheel
+        twine upload --username __token__ dist/*.whl
+
+    # Only do this on Windows to prevent file conflicts
+    - name: Build and publish source distribution
+      if: startsWith(matrix.os, 'windows')
+      env:
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        twine upload --username __token__ --repository-url https://test.pypi.org/legacy/ dist/*.tar.gz
```
TrueBrain: this repository url intended?

.github/workflows/testing.yml
```diff
@@ -0,0 +1,62 @@
...
+        pip install -e .
+
+    # Just check we can successfully build it
+    - name: Build standalone executable
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        pip install pyinstaller
+        pyinstaller ./nmlc.spec
+
+    # TODO: Add black/flake8 tests
```
TrueBrain, line 62: For another PR?

LordAro: That was the intent (will require quite a few code changes)

TrueBrain: I meant: please remove the comment and do this in another PR :P

.github/workflows/release.yml
LordAro: Not sure what you mean? All I know is that works on my repo (see release link in main body)

andythenorth:
Merging, as TB has approved.

PR #94, opened 2020-04-20
Change: Convert some code flows to context managers. (Supersedes #68)

Xaroth:
This extends on and supersedes #68 in that it replaces some open/close flows with context managers, making it slightly more resilient against exceptions and reducing the amount of file handles open.


andythenorth:
Confirmed that this resolves the open file limit issue in macOS.  

NewGRF otherwise compiles as expected.

I didn't try testing with a known-bad palette in an image, going to assume that still works :)

Issue #76, opened 2019-12-21
nfo output is slow with pypy


andythenorth:
nmlc is generally faster with pypy than python 3.8

However output to nfo is specifically about 5x slower with pypy compared to python 3.8.  

This appears to be due to poor performance of StringIO, which is a known issue with pypy: https://bitbucket.org/pypy/pypy/pull-requests/132/speed-up-pickledumps-fixes-issue-979

I created a test substituting StringIO with simple list append / join, which performs much faster
https://github.com/andythenorth/nml/commit/b2c9338d19fc7b1ca8e72a056bc25ba9abe232b9

For Iron Horse compile the difference is ~10s with pypy vs ~2s with python 3.8.  In the slower version the extra 8s represents ~25% of total compile time for the grf, so it is not insignificant.

My patch is not a viable candidate for merging, but this would be a helpful issue to fix.

andythenorth:
FLHerne has a patch which substantially reduces output time with pypy in Iron Horse and FIRS compiles http://www.flherne.uk/files/nicer-quick-output.diff

andythenorth:
I ran some tests against http://www.flherne.uk/files/nicer-quick-output.diff using [Iron Horse](https://github.com/andythenorth/iron-horse/commit/889de44cc3d6a82030603524f095444e66d0dedf and nml at rev 660722b680fd6a6c5b421d8eabbf36fcc82635ba
) 

Timings are purely for 'Writing output' step, tested over a couple of runs, with primed caches.

**Python 3.8**
 
Without patch: ~1.5s
With patch: ~1.7s - but also note there's an ignored Exception "FileNotFoundError: [Errno 2] No such file or directory: './generated/iron-horse.nfo.e5kf2jss.nmltemp'"

**PyPy"

Without patch: ~6.0s
With patch: ~1.0s

FLHerne:
The slow performance of StringIO is fixed in PyPy 7.3.1.

PR #98, opened 2020-04-21
Change: Use PyPy's StringBuilder to speed up the use of StringIO.

Xaroth:
This is an alternative to #76 .

Inspiration was found from https://bitbucket.org/pypy/pypy/pull-requests/132/speed-up-pickledumps-fixes-issue-979/diff , where a similar fix was applied to speed up pickle.dumps for PyPy.

This only affects PyPy, normal CPython versions will still use `io.StringIO`, therefore not incurring any performance loss.

### DIFF
--- a/nml/output_base.py
+++ b/nml/output_base.py
@@ -16,7 +16,9 @@ with NML; if not, write to the Free Software Foundation, Inc.,
 """
 Abstract base classes that implements common functionality for output classes
 """
-import array, io
+import array
+import io
+from nml.util import StringIO
 
 class OutputBase:
     """
@@ -286,7 +288,7 @@ class TextOutputBase(OutputBase):
         OutputBase.__init__(self, filename)
 
     def open(self):
-        self.file = io.StringIO()
+        self.file = StringIO()
 
 
 class BinaryOutputBase(SpriteOutputBase):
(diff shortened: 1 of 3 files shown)


andythenorth:
Some test results using this PR with PyPy 7.3.0

* about 40% faster for a newgrf compile compared to nml 9fdd4bbc8a8255f5108c801fc12bf9347be6f8f1
* the 'Writing nfo' output step is specifically faster by about 85%

Python 3.8 timings are unchanged, as expected.

FLHerne:
Much neater than my attempt!

That said, PyPy upstream have [fixed this in 7.3.1](https://doc.pypy.org/en/latest/whatsnew-pypy3-7.3.1.html):

> Improve performance of io.StringIO(). It should now be faster than CPython in common use cases.

(the changelog categories are confusing, but I've verified that the relevant commits appear in 7.3.1 and not 7.3.0).

That's done by [internally using a StringBuffer](https://bitbucket.org/pypy/pypy/branch/StringIO-perf) somewhat like this approach...

Testing on Linux with pypy 7.3.1, there is no measurable difference in performance with this patch; the present code is as fast with 7.3.1 as the patched version in 7.3.0 and earlier.

andythenorth:
I think we can rely on 7.3.1, so I'm going to close this - thanks for all the contributions!

PR #102, opened 2020-04-23
Fix #100: follow PEP440 when generating version

### LINKED ISSUE #100: Development version format is "invalid" to setuptools.
When running `setup.py` I get the following warning:

```/usr/lib/python3.8/site-packages/setuptools/dist.py:468: UserWarning: The version specified ('2020-04-20-g9fdd4bbc') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.```

[Link to PEP 440](https://www.python.org/dev/peps/pep-0440/).

It looks like the preferred version format would be something like `0.4.5.post87+g9fdd4bbc`

### PR DESCRIPTION
**Pull request by glx22**
Previous, and invalid, version format was `<year>-<month>-<day>[-<branch>]-g<changeset>[M]`.
I replaced it with `<next nml version>.dev<year><month><day>+[branch.]g<changeset>[m]`.
For releases it will still use `<tag>[m]`, and a modified tag source will be considered as invalid.

### DIFF
--- a/nml/version_info.py
+++ b/nml/version_info.py
@@ -93,8 +93,11 @@ def get_git_version(detailed = False):
         # Look for modifications
         try:
             modified  = (len(get_child_output(["git", "-C", path, "diff-index", "HEAD"], env=env)) > 0)
-            changeset = get_child_output(["git", "-C", path, "rev-parse", "--verify", "HEAD"], env=env)[0][:8]
             isodate   = get_child_output(["git", "-C", path, "show", "-s", "--pretty=%ci", "HEAD"], env=env)[0]
+            describe  = get_child_output(["git", "-C", path, "describe", "--tags", "--long"], env=env)[0].split('-')
+            tag = describe[0]
+            release = describe[1] == "0"
+            changeset = describe[2]
         except OSError as e:
             print("Git checkout found but cannot determine its version. Error({0}): {1}".format(e.errno, e.strerror))
             return version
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: nml/version_info.py [hunk 1]
```diff
@@ -121,13 +121,14 @@ def get_git_version(detailed = False):
         if tag:
             version = tag
             str_tag = tag
-        elif branch == "master":
-            version = isodate + "-g" + changeset
         else:
-            version = isodate + "-" + branch + "-g" + changeset
+            version = "0.5.0.dev" + isodate.replace("-", "") + "+"
```
FLHerne, line 125: Is hardcoding the 'next' version here a good idea? I feel it's guaranteed that someone will forget to update it eventually.

Might be safer to have it use the most recent tag and `.post`.

planetmaker: It's difficult to use the most recent tag. Traditionally NML created branches for the current stable version(s) starting with the major release of it and continued development on master. So the last tag might actually also mis-represent the actual version. Not sure there is a good solution to this problem.

 So changing the next major version to be released in master is not worst of ideas and only infrequently needs doing.

glx22: And current master can be incompatible with most recent tag so it would be misleading.

TrueBrain: I assume what @FLHerne was refering to: find the first tag from HEAD, and add ".post0" behind that. This works for branches, and doesn't care that master can be incompatible.

I guess you can read "recent" in several ways. But in the way I explain, it should work just fine :)

Otherwise this piece of code is weird, and will result in annoying problems. Why else do this git-detection at all, and why not just hardcode the version in the first place? :D (either do it all one way, or all the other, I would say)

nml/version_info.py
```diff
@@ -91,10 +91,13 @@ def get_git_version(detailed = False):
             pass
 
         # Look for modifications
-        try:
+        try:
```
LordAro, line 94: Oh no, trailing whitespace!

PR #103, opened 2020-04-23
Blacken & Flake8 nml

LordAro:
Quite a vast amount of changes. The middle black formatting commit probably isn't worth looking at too closely, but the other two commits should be quite readable.

PR #104, opened 2020-04-25
Doc: Changelog for 0.5.0 release

andythenorth:
This isn't 100%, there may be late additions via #102 and #103.

Known issues:
* would benefit from consolidation, e.g. some of the 'fix' commits are for features only introduced in this release
* capitalisation of first word after ':' is inconsistent
* use of commit prefix is inconsistent, e.g. 'change' commits relate to build environment, don't affect nml end user
* order is currently by prefix, then alphabetical, rather than by related items from the user's point-of-view, maybe that should be changed?

Some of these we should fix, some aren't a big deal for this time :)
 
`TrueBrain "as it goes: learn from the past, live in the future"`

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -1,3 +1,116 @@
+0.5.0 (2020-04-26)
+------------------------------------------------------------------------
+
+NML 0.5.0 is a substantial release, covering multiple areas of the newgrf spec.  Thanks to everyone who helped.
+
+
+Industries
+----------
+NOTE TO AUTHORS: NML 0.5.0 removes many industry properties and vars.
+Industry sets using these properties or vars will *not* compile with nml 0.5.0 and need to be updated.
+All industry vars removed in 0.5.0 are replaced by equivalent new vars.
+Production callback (produce block) format also changed so produce blocks will need to be updated.
+We don't break NML backwards compatibility lightly, but this change was necessary to support 16 accepted and 16 produced cargos.
+Thanks.
+
+- Support for up to 16 accepted cargos and 16 produced cargos (available in OpenTTD 1.9.0 onwards)
+- New industry properties and vars
+  - cargo_types
+  - incoming_cargo_waiting
+  - produced_cargo_waiting
+  - this_month_production
+  - last_month_production
+  - this_month_transported
+  - last_month_transported
+  - transported_last_month_pct
+  - production_rate
+- Removed legacy industry properties and vars
+  - accept_cargo_types
+  - prod_cargo_types
+  - prod_multiplier
+  - input_multiplier_1
+  - input_multiplier_2
+  - input_multiplier_3
+  - production_rate_1
+  - production_rate_2
+  - waiting_cargo_1
+  - waiting_cargo_2
+  - waiting_cargo_3
+  - produced_cargo_waiting_1
+  - produced_cargo_waiting_2
+  - produced_this_month_1
+  - produced_this_month_2
+  - produced_last_month_1
+  - produced_last_month_2
+  - transported_this_month_1
+  - transported_this_month_2
+  - transported_last_month_1
+  - transported_last_month_2
+  - transported_last_month_pct_1
+  - transported_last_month_pct_2
+- New approach to tile cargo acceptance
+  - use INDTILE_FLAG_ACCEPT_ALL to accept all cargos industry accepts
+  - or use up to 16 cargos with accepted_cargos property
+- Production callback format changed - see https://newgrf-specs.tt-wiki.net/wiki/NML:Produce
+
+Industry docs: https://newgrf-specs.tt-wiki.net/wiki/NML:Industries
+industry tile docs: https://newgrf-specs.tt-wiki.net/wiki/NML:IndustryTiles
+
+Houses
+------
+- Support for up to 16 accepted cargos at houses
+
+Houses docs: https://newgrf-specs.tt-wiki.net/wiki/NML:Houses
+
+
+Roadtypes and Tramtypes
+-----------------------
+- Support for Roadtypes and Tramtypes (available in OpenTTD 1.10.0 onwards)
+
+Roadtypes docs; https://newgrf-specs.tt-wiki.net/wiki/NML:Roadtypes
+Tramtypes docs: https://newgrf-specs.tt-wiki.net/wiki/NML:Tramtypes
+
+
+Railtypes
+---------
+- Increase railtype ID range from 16 to 64, to support 64 railtypes
+- Railtype flag RAILTYPE_FLAG_HIDDEN for hiding a railtype from player construction menu
+- Railtype flags RAILTYPE_FLAG_ALLOW_90DEG and RAILTYPE_FLAG_DISALLOW_90DEG for 90 degree curves
+- Railtype flags RAILTYPE_FLAG_HIDDEN and RAILTYPE_FLAG_PRECOMBINED for precombined spriteset type
+
+Railtypes docs: https://newgrf-specs.tt-wiki.net/wiki/NML:Railtypes
+
+
+Procedures
+----------
+- Support for use of switches and random switches as procedures, reducing repetition of switches
+
+There are no docs for this at the time of release.
+
+
+Other Changes
+-------------
+- Feature: Add ANIM_TRIGGER_APT_AIRPLANE_LANDS (follows OpenTTD #7182)
+- Feature: Add station var nearby_tile_grfid (0x6A)
+- Feature: Add TTDPatchFlag 256_persistent_registers (0x80)
+- Feature: PUSH/POP_COLOUR codes for strings
+- Feature: Debug_print for GRF parameters
+- Feature: Allow PLY to generate parsing/lexing tables (improves performance); regeneration can be forced with nmlc -R
+- Feature: syntax highlighting support for Visual Studio Editor
+- Change: improved colouring of nmlc errors and warnings (where shell/terminal supports this)
+- Change: Make sprite encoder warnings more informative (#89)
+- Add: nmlc -D option to enable debug mode for parser
+- Fix #36: Global variable traffic_side was inconsistent with global parameter traffic_side
+- Fix #39: Add compatibility with >=pillow-7.0.0
+- Fix #52: Compatibility with python 3.8
+- Fix #57: GRF parameters could be given negative values
+- Fix: Add missing parameter in industry and airport tiles in RandomAction2
+- Fix: Properly clean up resource handles (eliminates 'too many open files' error on macOS with larger GRFs)
+- Codechange: Update nml version detection to support Git using similar approach to OpenTTD
+- Doc: Update example train nml to demonstrate use of sprite stack and vehicle length
+- Doc/Fix: Update the stated version requirements
+
+
 0.4.5 (2018-06-30)
 ------------------------------------------------------------------------
 - Feature: increase cargo ID range from 32 to 64, to support 64 cargos in OpenTTD 11ab3c4ea2f6a6d29efda8c9ba2af04194621ea7


LordAro:
#103 should absolutely not block anything, and i'd be tempted to say it shouldn't be in this release anyway, just on the (offchance) it could introduce some different behaviour

glx22:
#102 is not vital for a release as the tag will be used in this case.

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Feature: Support industry vars production_rate (0x70) and transported_last_month_pct (0x71)
+- Feature: Debug_print for GRF parameters
+- Feature: nmlc option to enable debug mode for parser
+- Feature: nmlc option to force regeneration of parser tables
+- Feature: Add TTDPatchFlag 256_persistent_registers (0x80)
+- Change: improved colouring of nmlc errors and warnings (where shell/terminal supports this)
+- Change: Make sprite encoder warnings more informative (#89)
+- Fix #36: Global variable traffic_side was inconsistent with global parameter traffic_side
+- Fix #39: Add compatibility with >=pillow-7.0.0
+- Fix #58: Loosen Pillow version constraint
```
LordAro, line 23: remove

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix #52: time.clock() has been removed in python 3.8
```
LordAro, line 24: remove

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix #57: Check coherency of GRF parameters limits
```
LordAro, line 25: fixes should describe what has been fixed, not what the fix was
```suggestion
- Fix #57: GRF parameters could be given negative values
```

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix 014febc: Wrong link to LICENSE file
```
LordAro, line 26: remove

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix 78cd84c: Restore 'setup.py install' usability on windows
```
LordAro, line 27: should not refer to commit hashes

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix: Add missing parameter in industry and airport tiles in RandomAction2
+- Fix: Convert some code flows to context managers (eliminates 'too many open files' error on macOS with larger grfs)
+- Fix: LOAD_TEMP() 'value' attribute is empty
+- Fix: Update various sprite counts
+- Codechange: Visual Studio Editor Script
```
LordAro, line 32: in general, all codechanges should be removed

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix #39: Add compatibility with >=pillow-7.0.0
+- Fix #58: Loosen Pillow version constraint
+- Fix #52: time.clock() has been removed in python 3.8
+- Fix #57: Check coherency of GRF parameters limits
+- Fix 014febc: Wrong link to LICENSE file
+- Fix 78cd84c: Restore 'setup.py install' usability on windows
+- Fix: Add missing parameter in industry and airport tiles in RandomAction2
+- Fix: Convert some code flows to context managers (eliminates 'too many open files' error on macOS with larger grfs)
+- Fix: LOAD_TEMP() 'value' attribute is empty
+- Fix: Update various sprite counts
```
LordAro, line 31: again, find why this was done, not what

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Feature: Increase railtype ID range from 16 to 64, to support 64 railtypes
+- Feature: Railtype flags RAILTYPE_FLAG_ALLOW_90DEG and RAILTYPE_FLAG_DISALLOW_90DEG for 90 degree curves
+- Feature: Railtype flags RAILTYPE_FLAG_HIDDEN and RAILTYPE_FLAG_PRECOMBINED for precombined spriteset type
+- Feature: Add ANIM_TRIGGER_APT_AIRPLANE_LANDS (follows OpenTTD #7182)
+- Feature: Add station var nearby_tile_grfid (0x6A)
+- Feature: PUSH/POP_COLOUR codes for strings
+- Feature: Support industry vars production_rate (0x70) and transported_last_month_pct (0x71)
+- Feature: Debug_print for GRF parameters
+- Feature: nmlc option to enable debug mode for parser
+- Feature: nmlc option to force regeneration of parser tables
```
FLHerne, line 17: This and "Codechange: Allow PLY to generate parsing/lexing tables (performance improvement)" should probably be rolled together. Something like:

> - Feature: Allow PLY to generate parsing/lexing tables (improves performance); regeneration can be forced with `nmlc -R`.

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Change: Make sprite encoder warnings more informative (#89)
+- Fix #36: Global variable traffic_side was inconsistent with global parameter traffic_side
+- Fix #39: Add compatibility with >=pillow-7.0.0
+- Fix #58: Loosen Pillow version constraint
+- Fix #52: time.clock() has been removed in python 3.8
+- Fix #57: Check coherency of GRF parameters limits
+- Fix 014febc: Wrong link to LICENSE file
+- Fix 78cd84c: Restore 'setup.py install' usability on windows
+- Fix: Add missing parameter in industry and airport tiles in RandomAction2
+- Fix: Convert some code flows to context managers (eliminates 'too many open files' error on macOS with larger grfs)
```
LordAro, line 29:
```suggestion
- Fix: Properly clean up resource handles (eliminates 'too many open files' error on macOS with larger GRFs)
```

docs/changelog.txt
LordAro, line 32: this should be a feature though

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Feature: Allow use of switches and random switches as procedures, reducing repetition of switches
+- Feature: Increase railtype ID range from 16 to 64, to support 64 railtypes
+- Feature: Railtype flags RAILTYPE_FLAG_ALLOW_90DEG and RAILTYPE_FLAG_DISALLOW_90DEG for 90 degree curves
+- Feature: Railtype flags RAILTYPE_FLAG_HIDDEN and RAILTYPE_FLAG_PRECOMBINED for precombined spriteset type
+- Feature: Add ANIM_TRIGGER_APT_AIRPLANE_LANDS (follows OpenTTD #7182)
+- Feature: Add station var nearby_tile_grfid (0x6A)
+- Feature: PUSH/POP_COLOUR codes for strings
+- Feature: Support industry vars production_rate (0x70) and transported_last_month_pct (0x71)
+- Feature: Debug_print for GRF parameters
+- Feature: nmlc option to enable debug mode for parser
```
FLHerne, line 16: Add (`-D`) to this line?

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Codechange: Allow PLY to generate parsing/lexing tables (performance improvement)
+- Codechange: Avoid duplicating rail/road/tramtype table code
+- Codechange: Update nml version detection to support Git using similar approach to OpenTTD
+- Codechange: Mechanical cleanups of Python code
+- Codechange: Remove cx_Freeze
+- Codechange: Rewrite setup.py
+- Codechange: Simplify Pillow imports
+- Codechange: Simplify Pillow version detection
+- Codechange: Move license file to root folder
+- Doc: Update README
```
LordAro, line 44: remove

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Doc: 0.3 changed 'shorten_vehicle' to 'length' - this wasn't updated in one place in train example nml
+- Doc: Add example train wagon showing sprite stack use
+- Doc: Fix 404 on pillow install instructions link
```
LordAro, line 47: remove

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Feature: Houses now support up to 16 accepted cargos
+- Feature: Support for RoadTypes (a.k.a NRT)
+- Feature: Allow use of switches and random switches as procedures, reducing repetition of switches
+- Feature: Increase railtype ID range from 16 to 64, to support 64 railtypes
+- Feature: Railtype flags RAILTYPE_FLAG_ALLOW_90DEG and RAILTYPE_FLAG_DISALLOW_90DEG for 90 degree curves
+- Feature: Railtype flags RAILTYPE_FLAG_HIDDEN and RAILTYPE_FLAG_PRECOMBINED for precombined spriteset type
+- Feature: Add ANIM_TRIGGER_APT_AIRPLANE_LANDS (follows OpenTTD #7182)
+- Feature: Add station var nearby_tile_grfid (0x6A)
+- Feature: PUSH/POP_COLOUR codes for strings
+- Feature: Support industry vars production_rate (0x70) and transported_last_month_pct (0x71)
```
FLHerne, line 14: Reorder so this is with the 16-cargo industry change above? IIRC this was needed to support production rates for all those cargoes.

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Fix 78cd84c: Restore 'setup.py install' usability on windows
+- Fix: Add missing parameter in industry and airport tiles in RandomAction2
+- Fix: Convert some code flows to context managers (eliminates 'too many open files' error on macOS with larger grfs)
+- Fix: LOAD_TEMP() 'value' attribute is empty
+- Fix: Update various sprite counts
+- Codechange: Visual Studio Editor Script
+- Codechange: Install_requires in setup.py
+- Codechange: pyinstaller spec file
+- Codechange: Allow PLY to generate parsing/lexing tables (performance improvement)
+- Codechange: Avoid duplicating rail/road/tramtype table code
```
FLHerne, line 36: That code didn't exist in the last release, so not really a change for these purposes.

docs/changelog.txt
```diff
@@ -1,3 +1,54 @@
...
+- Codechange: Rewrite setup.py
+- Codechange: Simplify Pillow imports
+- Codechange: Simplify Pillow version detection
+- Codechange: Move license file to root folder
+- Doc: Update README
+- Doc: 0.3 changed 'shorten_vehicle' to 'length' - this wasn't updated in one place in train example nml
+- Doc: Add example train wagon showing sprite stack use
+- Doc: Fix 404 on pillow install instructions link
+- Doc/Fix: Update the stated version requirements
+- Cleanup: Industry vars production_rate_1 (0x8E) and production_rate_2 (0x8F) are deprecated and replaced by new parametric var production_rate (0x70)
```
FLHerne, line 49: Do the old ones still work? If not this should be much higher.

docs/changelog.txt
LordAro: I'd probably say it should be an "Add: " as well, it's not exactly a "feature"

docs/changelog.txt
LordAro: Indeed, if it's not an actual removal, it's not a "cleanup", rather a "change"

docs/changelog.txt
FLHerne: Why? This incompatibility was a specific reason the Debian packager was asking for a prompt release of 0.5.0.

LordAro: That's true. Perhaps
```suggestion
- Fix #52: Compatibility with python 3.8
```

docs/changelog.txt
FLHerne: Yeah, end-users don't care, this only makes nml developers' life easier.

docs/changelog.txt
andythenorth: It's to support newly added UI + tram sprites for NRT, it's not really relevant to end user

docs/changelog.txt
FLHerne: Both of the fixes above are for problems introduced since 0.4.5, so they should be removed from the changelog.

docs/changelog.txt
andythenorth: +1 this will cause some people's compiles to fail;  I really think this has to be out-of-order and follow directly from the 'Feature: production_rate 0x70 blah blah' line.

docs/changelog.txt
```diff
@@ -1,3 +1,46 @@
+0.5.0 (2020-04-26)
+------------------------------------------------------------------------
+
```
LordAro, line 3: extra newline

docs/changelog.txt
```diff
@@ -1,3 +1,46 @@
...
+- Fix #39: Add compatibility with >=pillow-7.0.0
+- Fix #52: time.clock() has been removed in python 3.8
+- Fix #57: Check coherency of GRF parameters limits
+- Fix: Add missing parameter in industry and airport tiles in RandomAction2
+- Fix: Convert some code flows to context managers (eliminates 'too many open files' error on macOS with larger grfs)
+- Fix: LOAD_TEMP() 'value' attribute is empty
+- Codechange: Rewrite setup.py
+- Codechange: Install_requires in setup.py
+- Codechange: pyinstaller spec file
+- Codechange: Update nml version detection to support Git using similar approach to OpenTTD
```
LordAro, line 33: keep this codechange, i don't see the need in keeping any others

docs/changelog.txt
```diff
@@ -1,3 +1,37 @@
+0.5.0 (2020-04-26)
+------------------------------------------------------------------------
+- Feature: Support for RoadTypes (a.k.a NRT)
+- Feature: Industries now support up to 16 accepted cargos and 16 produced cargos
```
FLHerne, line 4: Looking at the commit, this isn't nearly enough information or even an accurate summary.

It doesn't only *add* support for more cargos, it removes the previous syntax and properties in a way that will break all industry-related grfs.

The changelog needs to say that, and either list the required changes or link to somewhere that does.

(IMO such a big change should really have had a deprecation period, but too late now),

nielsmh: > (IMO such a big change should really have had a deprecation period, but too late now),

I initially left support for the old syntax in, but then was told that the NML way is to remove all traces of the old way of doing things when a new and better way is introduced.

Agree that the changelog should at least warn that the syntax has changed and all projects that define industries need to be updated, or stay on the old version.

LordAro:
The issue is that these are now release notes, rather than a changelog. There's no particular issue to doing it like this, but it's not how we've done any other changelog... ever. Consider splitting it out into a separate RELEASE_NOTES.md file?

frosch123:
I like the new way.

The old changelog contains stuff like "- Update: Changelog", which is mind-boggling stupid. Link to the VCS log in that case.

PR #108, opened 2020-04-27
Codechange: [Actions] Upload standalone executable to GitHub release

glx22:
I merged it into my [master](https://github.com/glx22/nml/commits/master) (with a temp commit to disable PyPi stuff. Then I created a [release](https://github.com/glx22/nml/releases/tag/test) to trigger it.

Seems to work as intended. Actions run log is [here](https://github.com/glx22/nml/runs/620714617?check_suite_focus=true).

Hmm I should probably rebase and update the `README` too. And maybe add the changelog in the zip.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,8 @@
 name: Release
 
 on:
-  push:
-    tags:
-    - '*'
+  release:
+    types: [published]
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


andythenorth:
We're going to need 0.5.1 soon for #109 so that might be a good live test for this?

LordAro:
Merged into #113

Issue #109, opened 2020-04-27
Roadtypes and Tramtypes size is 63 not 15 or 64


andythenorth:
*Current*

- OpenTTD provides for a shared pool of 63 Roadtypes and Tramtypes ("frosch: one id is used for 'no road/tram'")
- nml sizes Roadtype and Tramtype at 15 
  - https://github.com/OpenTTD/nml/blob/master/nml/actions/action0.py#L207
- NewGRF docs size Roadtypes and Tramtypes at 64
  - https://newgrf-specs.tt-wiki.net/wiki/Action0/Roadtypes
  - https://newgrf-specs.tt-wiki.net/wiki/NML:Roadtypes

This is a blocking issue and should be an easy fix.

*Desired*

- confirm that the OpenTTD pool is 63
- update nml to 63
- update docs to 63
- release nml 0.5.1

FLHerne:
@frosch123 - are you sure they're shared?
According to the docs in landscape.html:

> Bits 5..0: Road roadtype, 0x3F for no road.
> Bits 11..6: Tram roadtype, 0x3F for no tram.

That's enough bits for each to have 64 IDs (or 63 and 'none'?) 

@PeterN must know...

PeterN:
They're shared.

glx22:
https://github.com/OpenTTD/OpenTTD/blob/1f1345de098294a4744981d0043512569a35102a/src/road.h#L197-L208
https://github.com/OpenTTD/OpenTTD/blob/1f1345de098294a4744981d0043512569a35102a/src/road_type.h#L32-L42

andythenorth:
NFO and NML wiki pages updated to correct limit of 63.

andythenorth:
Fixed by #110 and wiki docs changes, thanks all.

PR #110, opened 2020-04-27
Fix: Correct limit of available road/tramtypes.

FLHerne:
According to the NFO spec, there can be up to 63 each of road/tramtypes.

OpenTTD also requires that the *combined* total of road and tram types be
 no more than 63, and that a label used for a roadtype cannot be used for
 a tramtype or vice versa.

This directly contradicts the NFO specification, and nmlc will not
 check for, nor enforce, these additional limitations.

### DIFF
--- a/nml/actions/action0.py
+++ b/nml/actions/action0.py
@@ -204,8 +204,8 @@ used_ids = [
     BlockAllocation(  0,    255, "Object"),
     BlockAllocation(  0,     63, "Railtype"),
     BlockAllocation(  0,    255, "Airport Tile"),
-    BlockAllocation(  0,     15, "Roadtype"),
-    BlockAllocation(  0,     15, "Tramtype"),
+    BlockAllocation(  0,     62, "Roadtype"),
+    BlockAllocation(  0,     62, "Tramtype"),
 ]
 
 def print_stats():


andythenorth:
NFO and NML wiki pages updated to correct limit of 63.

andythenorth:
Addresses the nmlc part of #109

FLHerne:
Based on @frosch123's recommendations on IRC.

Paging @andythenorth and @PeterN in case they have different ideas.

FLHerne:
IMO it was a mistake to merge this without further discussion; we still need to decide if nmlc or OTTD needs 'fixing' for the further issues.

andythenorth:
Ok, I'm +1 it's probably worth getting NRT right after so long and so much work by so many people :)

Full declaration, there's some burnout around NRT, me and probably other people too.
It's a very long time since I started a patch to simply remove catenary from tramway with a flag.  That grew rather quickly into NRT.

Although it's wholly irrelevant to the PR, this ticket is as good a place as any to put the NRT history, which includes.
* using GitHub and git (badly, with lots of merges) when the main project was in SVN
* having the binaries built by the official compile farm, which was good and bad (made it trivially 
available to players for testing, but also set expectations that it HAD to be be included in trunk)
* the clown shoes NML patch I wrote, again in Git (with bad merges), not the NML mercurial repo
* the happy background noise of peanuts being thrown that the spec is all wrong
* the sense of obligation from quite a lot of authors putting quite a lot of work into writing NRT newgrfs
* the desire to ship something big and prove that vanilla OpenTTD isn't dead
* the realisation that there's nowhere to hide from doing the docs, oof

So yeah, it's been emotional :)

I don't regret NRT overall, but I definitely have regrets, and would kind of like it to go away as my problem, I guess that's not how it works :)

Wolfolo:
I'm +1 for fixing NRT if needed (I don't know a bit of NML, so I can't talk about that), I would have followed precisely the specs if I didn't get confused about the constant changes, opinions, IRC talk, busy with work, my inability to get things right and the difficulty to understand that it was no more my (and Andy's) little project.
I'm also +1 to talk about what fix it need if OTTD-side, just poke me.

PR #113, opened 2020-05-01
Rewrite release workflow

LordAro:
Split up the release workflow into separate jobs, as trying to keep them together was just making things difficult to read.
Now builds source tarball with MacOS to get that LF line ending goodness

Seems to still work - https://github.com/LordAro/nml/actions/runs/93337764 - they all failed, but that's just because they don't have permission to upload to PyPI :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,37 +5,53 @@ on:
     types: [published]
 
 jobs:
-  release:
-    name: Python 3.x on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, macOS-latest, windows-2016]
+  release-ubuntu:
+    name: Ubuntu release
+    runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
 
     - uses: actions/cache@v1
-      if: startsWith(matrix.os, 'ubuntu')
       with:
         path: ~/.cache/pip
         key: ${{ runner.os }}-pip-${{ hashFiles('**/setup.py') }}
         restore-keys: |
           ${{ runner.os }}-pip-
 
-    - uses: actions/cache@v1
-      if: startsWith(matrix.os, 'macOS')
+    - name: Set up Python
+      uses: actions/setup-python@v1
       with:
-        path: ~/Library/Caches/pip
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -6,35 +6,92 @@ on:
...
 
 jobs:
-  release:
-    name: Python 3.x on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+  release-ubuntu:
+    name: Ubuntu release
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
```
matthijskooijman, line 13: I thikn fail-fast (and all of strategy) is no longer needed, fail-fast is only between elements of the matrix in a single job, I believe. Same for the other jobs.

.github/workflows/release.yml
```diff
@@ -6,35 +6,92 @@ on:
...
+        python-version: '3.x'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+        pip install -e .
+
+    # Only publish source package on MacOS to prevent PyPI file conflicts
+    # and so we get LF line endings (useful for debian source packages)
```
matthijskooijman, line 76: Maybe say here "Only publish source package on one OS, to prevent PyPI file conflicts. Use MacOS rather than Windows to get LF rather than CRLF line endings (useful for Debian source packages), and MacOS rather than Linux, since the latter has a different build process.

README.md
```diff
@@ -140,6 +140,7 @@ For large releases and/or if there are deprecations or nml syntax changes, provi
 Example: https://github.com/OpenTTD/nml/blob/master/docs/changelog.txt
 4. Publish a new release using the release tool in the GitHub project: https://github.com/OpenTTD/nml/releases/new
 5. GitHub Actions will build the release and publish to PyPI (the Python package index).
+Download the source tarball, then upload it to the GitHub release (helpful for debian package maintainers).
```
matthijskooijman, line 143: Why not let the workflow do this directly, just like the win64 build is already uploaded as an artifact?

LordAro: See the step below - it's not currently done by the github workflow ;)

LordAro: Doing so is on the todo list though...

glx22: We could probably merge #108 and this PR together

matthijskooijman: Oh, seems this needs #108 to really directly upload, but I guess that would be even nicer?

PR #114, opened 2020-05-01
Fix: tags could contain -RCx

glx22:


nml/version_info.py
```diff
@@ -96,8 +63,10 @@ def get_git_version(detailed = False):
             isodate   = get_child_output(["git", "-C", path, "show", "-s", "--pretty=%ci", "HEAD"], env=env)[0]
             describe  = get_child_output(["git", "-C", path, "describe", "--tags", "--long"], env=env)[0].split('-')
             tag = describe[0]
-            release = describe[1] == "0"
-            changeset = describe[2]
+            if describe[1] != describe[-2]:
+                tag += describe[1].lower()
```
TrueBrain, line 67: This is a bit weird. Why not just: `tag = "-".join(describe[0:-2])`?

nml/version_info.py
```diff
@@ -112,7 +79,7 @@ def get_git_version(detailed = False):
             branch = "master"
 
         # Compose the actual version string
-        version = tag
+        version = tag.replace("-", "").lower()
```
TrueBrain, line 82: Personally I would add a comment here what my intention is; just for future us.

PR #117, opened 2020-05-04
Fix #116: towname used bits could be the same for sub-parts

### LINKED ISSUE #116: TownName Resonance Issue
*Please let me know if this issue should be filed directly against the main OpenTTD repo rather than here.*

With the following code:

```
// lang/english.lng

##grflangid 00

STR_GRF_NAME						:Test Town Names
STR_GRF_DESCRIPTION					:Test TownName GRF{}{BLACK}Version: {SILVER}1 [2020-05-02]

STR_GAME_OPTIONS_TOWN_NAME_Test	:Test

STR_OPENTTD_VERSION                 :ERROR. Requires a newer version of OpenTTD. Please upgrade.
```

and

```
// testname.nml
grf {
    grfid: "STNM";
    name: string(STR_GRF_NAME);
    desc: string(STR_GRF_DESCRIPTION);
    version: 1;
    min_compatible_version: 1;
}

if (version_openttd(1,2,0,22723) > openttd_version) {
    error(FATAL, REQUIRES_OPENTTD, string(STR_OPENTTD_VERSION));
}

// prefixes
town_names(prefixes) {
    {
        text("1", 1),
        text("2", 1),
        text("3", 1),
        text("4", 1),
        text("5", 1),
    }
}

town_names(bodies) {
    {
        text("A", 1),
        text("B", 1),
        text("C", 1),
        text("D", 1),
        text("E", 1),
    }
}

town_names(main) {
    styles: string(STR_GAME_OPTIONS_TOWN_NAME_TEST);

    {
        town_names(prefixes, 1),
    }
    {
        text("-", 1),
    }
    {
        town_names(bodies, 1),
    }
}
```

This will only produce 5 town names, on any size map: 1-A, 2-B, 3-C, 4-D, 5-E.

image

I know that OpenTTD will stop attempting to add towns if it fails to successfully create a new town name enough times, so I think what is happening is that the map engine is going down the same paths again and again and hitting the same 5 names, rather than generating all possible 25 names, and then giving up.

I would expect this to give all 25 names, or at least more than 5.

I'm reporting this in part because I expect that there may be other combinations that hit the "resonance".

glx22:


nml/actions/actionF.py
```diff
@@ -108,15 +108,19 @@ def prepare_output(self, sprite_num):
...
+        # Don't overlap with existing action F
+        if len(town_names_blocks) == 0: startbit = 0
+        else: startbit = max(block.free_bit for block in town_names_blocks.values())
+
         town_names_blocks[self.id_number] = self # Add self to the available blocks.
 
         # Ask descendants for the lowest available bit.
-        if len(blocks) == 0: startbit = 0 # No descendants, all bits are free.
-        else: startbit = max(town_names_blocks[block].free_bit for block in blocks)
+        if len(blocks) != 0: startbit = max(town_names_blocks[block].free_bit for block in blocks)
```
frosch123, line 118: They may again create overlapping bits.
Shouldn't it always use the result from line 112/113?

PR #118, opened 2020-05-06
Fix: Allow calculation for palette in spritelayout

glx22:
Fixes https://dev.openttdcoop.org/issues/8519

### DIFF
--- a/nml/actions/action2layout.py
+++ b/nml/actions/action2layout.py
@@ -313,10 +313,12 @@ class Action2LayoutSprite:
                 self.create_register(name, offset)
             return val
         else:
+            self.palette_from_action1 = False
             if isinstance(value, expression.ConstantNumeric):
                 generic.check_range(value.value, 0, (1 << 14) - 1, "palette", value.pos)
-            self.palette_from_action1 = False
-            return value
+                return value
+            self.create_register(name, value)
+            return expression.ConstantNumeric(0)
 
     def _validate_always_draw(self, name, value):
         if not isinstance(value, expression.ConstantNumeric):
--- a/nml/actions/action2layout.py
+++ b/nml/actions/action2layout.py
@@ -281,6 +281,8 @@ class Action2LayoutSprite:
             if isinstance(value, expression.ConstantNumeric):
                 generic.check_range(value.value, 0, (1 << 14) - 1, "sprite", value.pos)
                 return value
+            if value.supported_by_actionD(raise_error=False):
+                return value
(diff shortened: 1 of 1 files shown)


glx22:
Indeed there was a missing '<< 16', but I'm not even sure this code is used.

glx22:
It was used for `param[]` before my first commit, then it was unused, then I reintroduced its usage for both sprites and palettes.

nml/actions/action2layout.py
```diff
@@ -287,13 +287,18 @@ def _validate_sprite(self, name, value):
             self.sprite_from_action1 = True
             val, offset = self.resolve_spritegroup_ref(value)
             if offset is not None:
-                self.create_register(name, offset)
+                if offset.supported_by_actionD(False):
+                    val = expression.BinOp(nmlop.ADD, val, offset, value.pos)
+                else:
+                    self.create_register(name, offset)
```
glx22, line 293: Not sure about this change, I think it will break things because `1 | 0002 | Add offset to sprite, disable default usage of construction stage or railtype-offset` flag is not set if actionD is used

glx22: After more reading of the spec I decided it was definitely wrong.

nml/actions/action2layout.py
```diff
@@ -216,7 +216,7 @@ def get_sprite_number(self):
         if add_sprite:
             expr = expression.BinOp(nmlop.ADD, sprite, expr, sprite.pos)
         if add_palette:
-            expr = expression.BinOp(nmlop.ADD, palette, expr, sprite.pos)
+            expr = expression.BinOp(nmlop.ADD, expression.BinOp(nmlop.SHIFT_LEFT, palette, expression.ConstantNumeric(16, sprite.pos), sprite.pos), expr, sprite.pos)
         return expr.reduce()
```
FLHerne, line 220: The conditionals and constant arithmetic about `add_sprite` and `add_palette` can be removed; `BinOp.reduce()` does this anyway.

nml/actions/action2layout.py
```diff
@@ -294,6 +281,8 @@ def _validate_sprite(self, name, value):
             if isinstance(value, expression.ConstantNumeric):
                 generic.check_range(value.value, 0, (1 << 14) - 1, "sprite", value.pos)
                 return value
+            if value.supported_by_actionD(False):
```
LordAro, line 284: I'd suggest using a named parameter here (and below)

PR #119, opened 2020-05-07
Add: Builtin functions `roadtype()` and `tramtype()`.

FLHerne:
This only adds to the copy-pasting, but eh.


Yexo:
"Fix: add" reads funny. Shouldn't this be simply "Add: new 'roadtype()'.... builtins"?

PR #131, opened 2020-05-08
Add: [Actions] Include pypy3 on Ubuntu for regression tests.

FLHerne:
Not on macOS or Windows, because installing Pillow doesn't work.
Prebuilt wheels for the correct PyPy version aren't [always?] available,
 and GitHub's macOS/Windows runners don't have zlib installed so it
 can't be built locally.


LordAro:
Should add a comment explaining why pypy3 is excluded

PR #133, opened 2020-05-09
Doc: Remove broken code from industry example.

FLHerne:
The removed code triggered an infinite loop in the production callback,
 without really illustrating how to achieve something useful.

Add a link to the wiki, which has a working example of how to use
 `produce` blocks.


glx22:
Hmm maybe keep a simple but working `produce()` call as it helped to find a `--nml` issue

FLHerne:
Superseded by #138

PR #136, opened 2020-05-10
Exclude version update code (fixes #112) and dotfiles from release tarballs.

### LINKED ISSUE #112: setup.py overwrites nml/__version__.py

FLHerne:
Closes #115


setup.py
```diff
@@ -4,9 +4,14 @@
 
 from setuptools import setup, Extension, find_packages
 from setuptools.command.build_py import build_py
-from nml import version_info
-NML_VERSION = version_info.get_and_write_version()
 
+try:
+    from nml import version_update
+    NML_VERSION = version_update.get_and_write_version()
+except ImportError:
```
LordAro, line 11: in normal usage, version_info would be used, not version_update, so perhaps that should go first?

glx22: When you create a bundle you want to update the version first. When ran from a bundle it will silently fall back to version_info.

FLHerne: What @glx22 says.

matthijskooijman:
I'm a bit late to the party, but these changes look good to me! Given `setuptools_scm` wasn't a clear all-in-one solution, it could have been used just as as library to parse git tags (and maybe PKG_INFO files), but since we already had code for that, probably better to not add one more dependency. Thanks for working on this!

PR #138, opened 2020-05-10
Fix: Make the industry example do something, rather than crashing

FLHerne:
Previously the produce() was an infinite loop.


andythenorth:
This looks correct to me, but I haven't tried compiling it and testing it in a test game (might be overkill).

PR #143, opened 2020-05-15
Add: Script to format the changelog for TT-Forums

FLHerne:
@andythenorth finds it annoying to do by hand.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -145,8 +145,7 @@ Other Changes
 - Fix: String encoding with cases was non-deterministic and resulted in differing md5sums for the NewGRF.
 - Fix: Rearrange order of string id assignment to keep compatibility.
 - Fix: When some IDs depleted, an internal NML error was raised instead of the intended error message.
-- Fix: most_common_refit was reading the wrong bits in var 42; read the write bits and rename it to most_common_cargo_type
-       as it's more accurate; also add most_common_cargo_subtype which also reads var 42 and was previously missing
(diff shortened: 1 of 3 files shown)


docs/changelog.txt
```diff
@@ -145,8 +145,7 @@ Other Changes
 - Fix: String encoding with cases was non-deterministic and resulted in differing md5sums for the NewGRF.
 - Fix: Rearrange order of string id assignment to keep compatibility.
 - Fix: When some IDs depleted, an internal NML error was raised instead of the intended error message.
-- Fix: most_common_refit was reading the wrong bits in var 42; read the write bits and rename it to most_common_cargo_type
-       as it's more accurate; also add most_common_cargo_subtype which also reads var 42 and was previously missing
+- Fix: most_common_refit was reading the wrong bits in var 42; read the write bits and rename it to most_common_cargo_type as it's more accurate; also add most_common_cargo_subtype which also reads var 42 and was previously missing
```
FLHerne, line 148: I removed the hard line-wrapping because the rest of the file doesn't use it, and I couldn't be bothered to handle it in the script.

TrueBrain:
Do such script really belong in these repositories? Where does that end? And I guess it also works with OpenTTD; so do we add our own version there?

In my opinion, mostly only 1 person uses these kind of scripts and know how to use it, and the repository becomes a personal storage for such scripts :)

In result, I am always a bit reluctant about these things. Isn't it better to place this either in a gist or in a separate repositories, with a collection of more of those scripts which are surrounding OpenTTD projects? I know for example we have language backport scripts, BaNaNaS scripts, ...

LordAro:
It's part of the release process, so I'd say it probably has a place in a proper repo somewhere. Perhaps in OTTD repo itself, with some other scripts (like language backporting)

FLHerne:
Closing. If there's demand we can find somewhere to put it.

PR #150, opened 2020-05-23
Fix #147: rounding issues

### LINKED ISSUE #147: Metric speed values are still slightly off
Using latest NML and OpenTTD there still seems to be something wrong with the metric speed units as mentioned in issue #74 and supposedly fixed in #75 . MPH works fine but KM/H values appear to be rounded wrong in OpenTTD despite being defined correctly in the NML code.

Unnamed, 01_01_1990#2

To check for yourself, load RattRoads (available on Bananas) and set the grf parameter "speed units" to km/h
Make sure your game setting "speed units" is also set to km/h and start a game
You should see the same result as shown above
Obviously the speeds should be 50 km/h, 60 km/h, 90 km/h, etc. but for some reason they are slightly off.

I also see a similar effect with m/s (but who really uses that anyway?)

FLHerne:


FLHerne:
@frosch123 -- you made the opposite change to the first commit before merging #75 , so I should probably ask why you think it's wrong. ;-)

FLHerne:
Hm, now the `round()` is redundant. I think it's that bit that's wrong.

FLHerne:
Updated -- I still don't really understand what the OpenTTD code does, but now I'm confident the Python version does the same thing...

nml/actions/action0properties.py
```diff
@@ -255,8 +255,11 @@ def cargo_list(value, max_num_cargos):
     'name'               : {'num': -1, 'string': None},
 }
 
-def ottd_display_speed(value, divisor, unit):
-    return (round(value.value / divisor * 10 * unit.ottd_mul) >> unit.ottd_shift) // 16
+def ottd_display_speed(value, mul, div, unit):
+    # Convert value to km/h-ish.
+    kmh_ish = value.value * mul // div
```
frosch123, line 260: While this now matches OTTD's GetDisplaySpeed, it throws away the fractional part of the speed.
You can define a road vehicles that has a max speed of 7 units (=3.5 km-ish/h), and the movement code will use that speed. OTTD will round it for display to 3 km-ish/h though.

So, ideally it should be possible to define vehicles in NML with a speeds of 3, 3.5 and 4 km/h, which have different speed in-game, and the 3 and 4 is also shown in the vehicle lists.

FLHerne: This function is only used to clamp the internal value calculated elsewhere, by incrementing/decrementing it until the display speed matches the input. Fractional values that display as the correct integer should be unaffected.

LordAro (approved):

PR #151, opened 2020-05-23
Add: Ignore unreferenced named parameters

glx22:
When a named parameter is defined, it consumes an actionD register, even if it's never accessed later.

This PR solves that by reference counting all identifiers, and assign an actionD register only if it's needed.

The check is not recursive, a named parameter referenced only by an unreferenced one won't be ignored. Same for write only named parameters written multiple times. But I think these cases should not be common.

### DIFF
--- a/nml/actions/actionD.py
+++ b/nml/actions/actionD.py
@@ -94,6 +94,9 @@ class ParameterAssignment(base_statement.BaseStatement):
             if not self.param.can_assign():
                 raise generic.ScriptError("Trying to assign a value to the read-only variable '{}'".format(self.param.name), self.param.pos)
         elif isinstance(self.param, expression.Identifier):
+            if global_constants.identifier_refcount[self.param.value] == 0:
+                generic.print_warning("Named parameter '{}' is not referenced, ignoring.".format(self.param.value), self.param.pos)
+                return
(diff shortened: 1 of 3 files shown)


nml/actions/actionD.py
```diff
@@ -94,6 +94,9 @@ def pre_process(self):
             if not self.param.can_assign():
                 raise generic.ScriptError("Trying to assign a value to the read-only variable '{}'".format(self.param.name), self.param.pos)
         elif isinstance(self.param, expression.Identifier):
+            if global_constants.identifier_refcount[self.param.value] == 0:
+                generic.print_warning("Named parameter '{}' is not referenced, ignoring.".format(self.param.value), self.param.pos)
+                return
```
FLHerne, line 99: I think it would be better to have the warning where it's actually handled, before `return []`.

That makes it easier to see what the code-path is doing, and avoids splitting the logic into two completely different places.

(I see the similar messages, but they're fatal ones that exit immediately with no handling elsewhere).

glx22:
Yeah it's a very basic optimisation, but it should handle all simple cases and free some actionD registers.

Anyway, please do not merge this PR before @Eddi-z can test it :)

PR #152, opened 2020-05-24
Codechange: Reintroduce 0.4 syntax for compatibility

glx22:
This allows to compile FIRS master branch with many warnings.

### DIFF
--- a/nml/actions/action0.py
+++ b/nml/actions/action0.py
@@ -433,12 +433,13 @@ def get_property_info_list(feature, name):
         raise generic.ScriptError("Setting properties for feature '{}' is not possible, no properties are defined.".format(general.feature_name(feature)), name.pos)
 
     if isinstance(name, expression.Identifier):
-        if not name.value in properties[feature]: raise generic.ScriptError("Unknown property name: " + name.value, name.pos)
-        prop_info_list = properties[feature][name.value]
+        prop_name = name.value
+        if prop_name not in properties[feature]:
+            raise generic.ScriptError("Unknown property name: " + prop_name, name.pos)
+        prop_info_list = properties[feature][prop_name]
         if not isinstance(prop_info_list, list): prop_info_list = [prop_info_list]
     elif isinstance(name, expression.ConstantNumeric):
-        for p in properties[feature]:
-            prop_info_list = properties[feature][p]
+        for prop_name, prop_info_list in properties[feature].items():
             if not isinstance(prop_info_list, list): prop_info_list = [prop_info_list]
             # Only non-compound properties may be set by number
             if len(prop_info_list) == 1 and 'num' in prop_info_list[0] and prop_info_list[0]['num'] == name.value:
@@ -448,6 +449,8 @@ def get_property_info_list(feature, name):
     else: assert False
 
     for prop_info in prop_info_list:
+        if 'replaced_by' in prop_info:
+            generic.print_warning("'{}' is deprecated, consider using '{}' instead".format(prop_name, prop_info['replaced_by']), name.pos)
         if 'warning' in prop_info:
             generic.print_warning(prop_info['warning'], name.pos)
     return prop_info_list
--- a/nml/actions/action0properties.py
+++ b/nml/actions/action0properties.py
@@ -709,6 +709,15 @@ def industry_layouts(value):
         layouts.append(tilelayout_names[name.value])
     return [IndustryLayoutProp(layouts)]
 
+def industry_prod_multiplier(value):
+    if not isinstance(value, Array) or len(value.values) > 2:
+        raise generic.ScriptError("Prod multiplier must be an array of up to two values", value.pos)
+    props = []
+    for i in range(0, 2):
+        val = value.values[i].reduce_constant() if i < len(value.values) else ConstantNumeric(0)
+        props.append(Action0Property(0x12 + i, val, 1))
+    return props
+
 class RandomSoundsProp(BaseAction0Property):
     def __init__(self, sound_list):
         self.sound_list = sound_list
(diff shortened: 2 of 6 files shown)


nml/actions/action0properties.py
```diff
@@ -849,7 +871,9 @@ def check_accept(acp):
     'prod_increase_msg'      : {'size': 2, 'num': 0x0D, 'string': 0xDC},
     'prod_decrease_msg'      : {'size': 2, 'num': 0x0E, 'string': 0xDC},
     'fund_cost_multiplier'   : {'size': 1, 'num': 0x0F},
-    'cargo_types'            : {'custom_function': industry_cargo_types}, # = prop 25+26+27+28 combined in one structure
+    'prod_cargo_types'       : {'size': 2, 'num': 0x10, 'value_function': lambda value: cargo_list(value, 2), 'warning': "Please consider replacing 'prod_cargo_types' with 'cargo_types'"},
```
FLHerne, line 874: I think copy-pasting all these warnings is a bit ugly. Something like `{..., 'replaced_by': "cargo_types"}` would be better, but `parse_var()` doesn't know the current name...

Fixing the TODO at action2var.py:951 would allow this, I will try to do that.

nml/actions/action0properties.py
```diff
@@ -859,11 +883,15 @@ def check_accept(acp):
...
     'new_ind_msg'            : {'size': 2, 'num': 0x1B, 'string': 0xDC},
+    'input_multiplier_1'     : {'custom_function': lambda value: industry_input_multiplier(value, 0x1C), 'warning': "Please consider replacing 'input_multiplier_1' with 'cargo_types'"},
+    'input_multiplier_2'     : {'custom_function': lambda value: industry_input_multiplier(value, 0x1D), 'warning': "Please consider replacing 'input_multiplier_2' with 'cargo_types'"},
+    'input_multiplier_3'     : {'custom_function': lambda value: industry_input_multiplier(value, 0x1E), 'warning': "Please consider replacing 'input_multiplier_3' with 'cargo_types'"},
     'name'                   : {'size': 2, 'num': 0x1F, 'string': 0xDC},
     'prospect_chance'        : {'size': 4, 'num': 0x20, 'unit_conversion': 0xFFFFFFFF},
     # prop 21, 22 (callback flags) are not set by user
     'remove_cost_multiplier' : {'size': 4, 'num': 0x23},
     'nearby_station_name'    : {'size': 2, 'num': 0x24, 'string': 0xDC},
+    'cargo_types'            : {'custom_function': industry_cargo_types}, # = prop 25+26+27+28 combined in one structure
```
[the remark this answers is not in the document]
glx22: I prefer to have the properties ordered by number :)

nml/actions/action0properties.py
glx22: Yeah for properties, name is known so it should be easy to add `replaced_by`. But for variables it's less easy.

nml/actions/action0.py
```diff
@@ -448,6 +448,8 @@ def get_property_info_list(feature, name):
     else: assert False
 
     for prop_info in prop_info_list:
+        if 'replaced_by' in prop_info:
+            generic.print_warning("Please consider replacing '{}' with '{}'".format(name.value, prop_info['replaced_by']), name.pos)
```
FLHerne, line 452: This doesn't work properly when 'name' is a number. I have a patch for that.

glx22: I applied your patch. And reworded other warnings too.

nml/actions/action2var_variables.py
```diff
@@ -506,7 +509,21 @@ def nearest_house_matching_criterion(name, args, pos, info):
...
+    'produced_cargo_waiting_1'     : {'var': 0x8A, 'start':  0, 'size': 16, 'replaced_by': 'produced_cargo_waiting'},
+    'produced_cargo_waiting_2'     : {'var': 0x8C, 'start':  0, 'size': 16, 'replaced_by': 'produced_cargo_waiting'},
+    'production_rate_1'            : {'var': 0x8E, 'start':  0, 'size':  8, 'replaced_by': 'production_rate'},
+    'production_rate_2'            : {'var': 0x8F, 'start':  0, 'size':  8, 'replaced_by': 'production_rate'},
     'production_level'             : {'var': 0x93, 'start':  0, 'size':  8},
+    'produced_this_month_1'        : {'var': 0x94, 'start':  0, 'size': 16, 'replaced_by': 'this_month_production'},
+    'produced_this_month_2'        : {'var': 0x96, 'start':  0, 'size': 16, 'replaced_by': 'this_month_production'},
+    'transported_this_month_1'     : {'var': 0x98, 'start':  0, 'size': 16, 'replaced_by': 'this_month_transported'},
+    'transported_this_month_2'     : {'var': 0x9A, 'start':  0, 'size': 16, 'replaced_by': 'this_month_transported'},
+    'transported_last_month_pct_1' : {'var': 0x9C, 'start':  0, 'size':  8, 'value_function': value_mul_div(101, 256), 'replaced_by': 'production_rate'},
```
FLHerne, line 521: Should be 'replaced_by': 'transported_last_month_pct' ?

andythenorth:
Thanks.  I guess this changes the precedent for future breaking changes.  :)

* Will the old syntax ever be removed?
* Should the docs 'removed' be changed to 'deprecated'? https://newgrf-specs.tt-wiki.net/wiki/NML:Industries

nielsmh:
The docs should change to "deprecated" yes, perhaps with a note that the syntax was entirely removed in a range of versions.

glx22:
I think the old syntax can stay as long as it's easy to support it. Anyway with all the warnings it outputs, it should be annoying enough to incite authors to switch.

PR #154, opened 2020-06-01
Add boolean constants

spnda:
I'm not sure if this is already a thing, but I don't think it is.
I feel like this is a neat addition but maybe there should be some checking that these are correctly used? They just expand to `0` and `1` respectively.

### DIFF
--- a/nml/actions/action0.py
+++ b/nml/actions/action0.py
@@ -204,8 +204,14 @@ used_ids = [
     BlockAllocation(  0,    255, "Object"),
     BlockAllocation(  0,     63, "Railtype"),
     BlockAllocation(  0,    255, "Airport Tile"),
+<<<<<<< HEAD
     BlockAllocation(  0,     62, "Roadtype"),
     BlockAllocation(  0,     62, "Tramtype"),
+=======
+    BlockAllocation(  0,     15, "Roadtype"),
+    BlockAllocation(  0,     15, "Tramtype"),
+    BlockAllocation(  0,    255, "RoadStop"),
+>>>>>>> Feature: NewGRF Roadstops (Feature 14)
 ]
 
 def print_stats():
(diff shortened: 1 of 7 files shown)


Yexo:
Do you have a particular use-case in mind for this?

spnda:
For example, the `dual-headed` for trains expects a 0 or 1, so false or true.

Yexo:
Fair enough, there are several other `vehicle_is_*` variables that are essentially bools. Then again, they're currently all documented as `0 | 1`. I see advantages in introducing a boolean type, but I'd rather do that properly instead of relying on integer values 0/1 disguising themselves as boolean.

That'd be a different approach than this PR is taking: the nml AST already has a Boolean type. Make use of that for `true` and `false` and then type-check all expressions including in switch-blocks to make sure those constants aren't mis-used.

frosch123:
Defining "true" as 0, and "false" as 1, is asking for disaster.

spnda:
> 
> 
> Fair enough, there are several other `vehicle_is_*` variables that are essentially bools. Then again, they're currently all documented as `0 | 1`. I see advantages in introducing a boolean type, but I'd rather do that properly instead of relying on integer values 0/1 disguising themselves as boolean.
> 
> That'd be a different approach than this PR is taking: the nml AST already has a Boolean type. Make use of that for `true` and `false` and then type-check all expressions including in switch-blocks to make sure those constants aren't mis-used.
Ok, I'll look into that.

> Defining "true" as 0, and "false" as 1, is asking for disaster.
ohh... oops

andythenorth:
Odd timing, I am looking at a case where True and False as constants might be useful.  

Using True and False for return values from a procedure might be more readable than the current 1 and 0 I'm using.

Not sure it's essential, and I could add them in my python compile as constants if I wanted.  I didn't do that so eh :)

spnda:
nevermind

PR #156, opened 2020-06-04
Fix: Procedure calls reference counting

glx22:


andythenorth:
Fixes my test case which was failing with master.

PR #164, opened 2020-10-02
Fix: Remove trailing whitespaces in NFO output

glx22:
Commit checker should now be happy.


LordAro:
It's a bit odd that it's always a single space for every single line.. why is it there in the first place?

glx22:
It's there because all writes add it, as it's simpler to always add a space in case there's something written after.

LordAro:
Add a comment so it's obvious that's the reason?

LordAro (approved):

PR #166, opened 2020-10-31
WIP: handle labels more readably

FLHerne:
Unfinished, but does this look sane?

See changes in `regression/expected`.


glx22:
Changes in `regression/expected` are correct. I like how ASCII values are printed, but full numerical values looked better with `dxXXXXXXXX`

FLHerne:
> Changes in `regression/expected` are correct. I like how ASCII values are printed, but full numerical values looked better with `dxXXXXXXXX`

You mean the ones like `"\12\34\56\78"`? That's where the input NML happened to use all escaped bytes in a string-literal, e.g. in this case
```// also override some other grf, this time not setting our own grfid explicitly
engine_override("\12\34\56\78");```

I think this is probably helpful because it makes it clearer where the value came from, and that adding some heuristic would be confusing.

This should never affect numeric values that weren't strings in the source code.

PR #168, opened 2020-11-04
CodeQL fixes and other misc. cleanups

FLHerne:


nml/actions/action2layout.py
```diff
@@ -353,13 +353,7 @@ def _validate_bounding_box(self, name, value):
...
 
     def _validate_hide_sprite(self, name, value):
-        value = expression.Not(value)
-        try:
-            value = value.reduce()
-        except:
-            pass
-        self.create_register(name, value)
-        return None
+        self.create_register(name, expression.Not(value).reduce())
```
glx22, line 356: Are we sure `reduce()` can't throw ?

FLHerne: It can raise a `ScriptError` if the expression doesn't return an integer. The only meaningful values for `hide_sprite` are 0 and 1, so in that case we should fail compilation.

(I did wonder why this try-except was added originally, but couldn't find an explanation)

FLHerne:
Note: the three "failing" CodeQL errors are the ones that I added suppression comments to, because of this bizarre design choice:
https://github.com/github/codeql/issues/4511

LordAro (approved):
LGTM

PR #169, opened 2020-11-09
Builtin functions refactoring

FLHerne:
Yay, more random cleanups.

Slightly RFC on the decorator bit; if this style makes sense there are probably more things to apply it to. Personally I think anything is better than these big tables.

### DIFF
--- a/nml/actions/action2var_variables.py
+++ b/nml/actions/action2var_variables.py
@@ -562,8 +562,7 @@ def industry_town_count(name, args, pos, info):
     return (args[0], extra_params)
 
 def industry_cargotype(name, args, pos, info):
-    from nml.expression.functioncall import builtin_cargotype
-    return (builtin_cargotype(name, args, pos), [])
+    return (expression.functioncall.builtin_resolve_typelabel(name, args, pos), [])
 
 varact2vars60x_industries = {
     'nearby_tile_industry_tile_id' : {'var': 0x60, 'start':  0, 'size': 16, 'param_function': unsigned_tile_offset},
--- a/nml/expression/functioncall.py
+++ b/nml/expression/functioncall.py
@@ -137,9 +137,36 @@ class GRMOp(Expression):
         return True
 
 
+function_table = {}
+
+
+def builtin(func):
+    """
+    Decorator that adds a function named `builtin_func` to the function table as `func`.
+    """
+    assert func.__name__.startswith("builtin_")
+    name = func.__name__[8:]  # Strip the "builtin_". str.removeprefix() is only added in py3.9.
+    function_table[name] = func
+    return func
+
+
+def builtins(*names):
+    """
+    Decorator that adds a function to the function table with one or more custom names.
+    """
+
+    def dec(func):
+        for name in names:
+            function_table[name] = func
+        return func
+
+    return dec
+
+
 # { Builtin functions
 
 
+@builtin
 def builtin_min(name, args, pos):
     """
     min(...) builtin function.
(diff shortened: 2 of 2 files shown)


nml/actions/action2var_variables.py
```diff
@@ -562,8 +562,14 @@ def industry_town_count(name, args, pos, info):
...
-    from nml.expression.functioncall import builtin_cargotype
-    return (builtin_cargotype(name, args, pos), [])
+    if len(args) != 1:
+        raise generic.ScriptError(name + "() must have 1 parameter", pos)
+    if not isinstance(args[0], expression.StringLiteral) or args[0].value not in global_constants.cargo_numbers:
+        raise generic.ScriptError(
+            "Parameter for {}() must be a string literal that is also in your cargo table", pos
+        )
+    num = expression.ConstantNumeric(global_constants.cargo_numbers[args[0].value])
+    return (num, [])
```
FLHerne, line 572: Ok, so this is *increasing* duplication...I'll try to find a better way to fix this.

(also, I just realized I forgot to format the format-string here...)

glx22: Can't import `function_table` from `functioncall` and use `function_table["cargotype"](name, args, pos)` ?

FLHerne: I couldn't then, because `name` wouldn't have been in the lookup table. Sorted now though.

LordAro (commented):
Not a huge fan of the "random" distinction between `@builtin` and `@builtin_names`, and with the naming of the functions. Could they be made more consistent? 

I'd probably go with always having a "builtin_" prefix to the function name and always using `@builtin_names`

nml/actions/action2var_variables.py
```diff
@@ -562,8 +562,7 @@ def industry_town_count(name, args, pos, info):
     return (args[0], extra_params)
 
 def industry_cargotype(name, args, pos, info):
-    from nml.expression.functioncall import builtin_cargotype
-    return (builtin_cargotype(name, args, pos), [])
+    return (expression.functioncall.builtin_resolve_typelabel(name, args, pos), [])
```
frosch123, line 565: builtin_resolve_typelabel would not need that "default to cargo" trick, if there was a 4th optional paramater for the function name, like
`
builtin_resolve_typelabel("cargo", args, pos, username=name)
`

FLHerne: > Not a huge fan of the "random" distinction between `@builtin` and `@builtin_names`, and with the naming of the functions. Could they be made more consistent?
> 
> I'd probably go with always having a "builtin_" prefix to the function name and always using `@builtin_names`

I changed `@builtin` to require and strip off the `builtin_` prefix, which removes most of the inconsistency.

Possibly I'm overthinking this and should just use `builtins("name")` for everything, but this way reduces the chance of copy-paste errors if people add new functions.

FLHerne:
> Not a huge fan of the "random" distinction between `@builtin` and `@builtin_names`, and with the naming of the functions. Could they be made more consistent?
> 
> I'd probably go with always having a "builtin_" prefix to the function name and always using `@builtin_names`

I changed `@builtin` to require and strip off the "builtin_" prefix, which removes most of the inconsistency. Perhaps I'm overthinking this, but it should reduce the chance of copy-paste errors if someone adds a new function.

nml/actions/action2var_variables.py
FLHerne: I really don't like this idea. The existing `name` parameter already has that meaning; passing in a fake value to manipulate the behaviour and then adding a new parameter for the real value seems like a far bigger hack. The default is going to exist somewhere, why not here?

(I updated the comment a bit)

frosch123 (approved):

PR #170, opened 2020-11-14
Switches optimisations improvements and fixes

glx22:
Before this rewrite, an optimised switch could output a `returns a constant, optimising` warning at each usage location, and a bogus `is not referenced, ignoring` warning. Now only one `returns a constant, optimising` happens.

There was also an bug in `is_read_only`.


LordAro (approved):
Code looks fine to me. Would be nice to have some extra tests, I suppose

PR #173, opened 2020-12-03
Fix: Access to persistent storage of towns

frosch123:
This PR adds a distinction between 'features' and 'scopes' (defined by feature+self/parent), instead of identifying both with interchangeable integers.
This fixes access to persistent storage to towns. (see #26, #28, https://www.tt-forums.net/viewtopic.php?p=1220012#p1220012 )


andythenorth:
\o/

Andrew350:
I'm not sure what it's worth, but I just tested this PR with my NewGRF, and I'm now able to successfully store a value via an industry and then read that value from a house to conditionally allow/disallow construction. So it seems to work for me 🙂

nml/actions/action2var_variables.py
```diff
@@ -779,30 +773,64 @@ def industry_cargotype(name, args, pos, info):
...
+        self.has_persistent_storage = has_persistent_storage
+
+
+class VarAct2Feature:
+    def __init__(self, self_scope, parent_scope):
+        self.self_scope = self_scope
+        self.parent_scope = parent_scope
+
+    def get_scope(self, var_range):
+        assert var_range in (0x89, 0x8A)
```
LordAro, line 790: asserts get removed in optimised builds (like Windows exe), and I think most others have been replaced with an explicit `raise AssertionError`. Not necessarily an issue, but something to consider?

nml/ast/switch.py
```diff
@@ -62,7 +62,9 @@ def pre_process(self):
             seen_names.add(param.value)
 
         feature = next(iter(self.feature_set))
-        var_feature = action2var.get_feature(self)  # Feature of the accessed variables
+        var_scope = action2var.get_scope(feature, self.var_range)
+        if var_scope is None:
+            raise generic.ScriptError("Requested scope not available for this feature.", self.pos)
```
LordAro, line 67: separate issue to deal with, but most ScriptErrors do not have a trailing '.':
```
$ grep -R "raise generic.ScriptError" nml | grep '[.]"[^"]*$' | wc -l
131
$ grep -R "raise generic.ScriptError" nml | grep '[^.]"[^"]*$' | wc -l
284
```

LordAro (commented):
Code looks fine, can't speak for its correctness

nml/actions/action2var_variables.py
frosch123: We removed all "assert False". If they were removed in release builds, they would open up new code paths.
A tool warned about that.
Regular asserts are fine.

nml/ast/switch.py
frosch123: Most messages in this file have a trailing colon.

glx22 (approved):
Seems correct, and it's a lot more cleaner than the old way.

PR #174, opened 2020-12-05
Add: make target to run black-formatter as it is used in PR checks.

frosch123:
I don't know how you invoke black. But there are quite some custom options for NML.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,13 @@
 MAKE?=make
 PYTHON?=/usr/bin/env python3
+BLACK_OPTIONS=-l 120 --exclude 'action2var_variables.py|action3_callbacks.py|generated'
 
-.PHONY: regression install bundle extensions clean
+.PHONY: regression test install extensions clean flake black
 
 regression: extensions
 	$(MAKE) -C regression
-test: regression
+
+test: regression flake
 
 install:
 	$(PYTHON) setup.py install
@@ -15,3 +17,10 @@ extensions:
 
 clean:
 	$(MAKE) -C regression clean
+
+flake:
+	$(PYTHON) -m black --check $(BLACK_OPTIONS) nml
+	$(PYTHON) -m flake8 nml
+
+black:
+	$(PYTHON) -m black $(BLACK_OPTIONS) nml
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ To install NML you'll need these 3rd party packages:
 - gcc (or possibly another c++ compiler).
   Needed to compile the cython version of the lz77 module for grf encoding.
 
+Running `make test` requires these code formatters and checkers. You don't need these to use NML, only to test or help improve it.
+
+- `flake8`
+  Minimal version is 3.7.
+- `black`
+
 ## 3) Installation
 
 The easiest way to install NML is by using pip:


FLHerne:
Comment as on IRC:

This should probably also add a `--check` run of black, and flake8, to `make test`.

Potentially we could then drop black and flake8 as separate GitHub actions because the main testing would run them, but then we'd lose inline notes from TrueBrain's action.

frosch123:
I added them to "make test".
But now "make test" depends on "flake8" and "black" being installed.

FLHerne:
I don't think that's an issue really; `black` at least you pretty much need to make any significant changes without upsetting the commit-checker.

Perhaps the people building distro packages will mind?

In principle we could have a test script that checks whether they're installed and warns but succeeds otherwise, but that seems TMWFTLB.

Makefile
```diff
@@ -15,3 +17,10 @@ extensions:
 
 clean:
 	$(MAKE) -C regression clean
+
+flake:
+	$(PYTHON) -m black -check $(BLACK_OPTIONS) nml
```
FLHerne, line 22: Should be `--check` with two `-`s.

FLHerne:
I guess this should add `black` and `flake8` to "Optional Dependencies" of README.md.

README.md
```diff
@@ -50,6 +50,12 @@ To install NML you'll need these 3rd party packages:
 - gcc (or possibly another c++ compiler).
   Needed to compile the cython version of the lz77 module for grf encoding.
 
+To improve NML and submit pull requests you'll need these code formaters and checkers.
```
FLHerne, line 53: [pedant] "formatters".

Also this seems a bit vague. Non-contributors who want to run `make test` (e.g. distro packagers) will also need these, and they aren't strictly *required* for making a PR.

 Perhaps something like:

> Running `make test` requires these code formatters and checkers. You don't need these to use NML, only to test or help improve it.

LordAro:
Not sure I like this being rolled into the `make test` rule. A separate "checkers" rule, perhaps?

Makefile
```diff
@@ -1,11 +1,13 @@
 MAKE?=make
 PYTHON?=/usr/bin/env python3
+BLACK_OPTIONS=-l 120 --exclude 'action2var_variables.py|action3_callbacks.py|generated'
```
TrueBrain, line 3: if you like, black options can also be put in `pyproject.toml`, see https://black.readthedocs.io/en/stable/pyproject_toml.html .
This makes sure that if someone executed `black`, it picks up the right rules too.
I have yet to apply this for all other OpenTTD projects :)

Makefile
```diff
@@ -15,3 +17,10 @@ extensions:
 
 clean:
 	$(MAKE) -C regression clean
+
+flake:
```
TrueBrain, line 21: My OCD cannot handle that `make flake` also executes black .. but using any other name is silly :P

TrueBrain (approved):
I think such quality of life improvements are always a good thing. You cannot execute tests often enough.

TrueBrain:
> 
> 
> Not sure I like this being rolled into the `make test` rule. A separate "check" rule, perhaps?

As much as a regression shows a runtime error, `flake` shows possible runtime errors. One can debate if `black` is really a test, but flake for sure is. And with having `flake` under `make test`, you might as well add `black` too, as .. why not! :D

frosch123:
Conclusion: noone likes pyproject.toml

PR #175, opened 2020-12-08
Fix mixing action features and variable scopes.

frosch123:


nml/ast/switch.py
```diff
@@ -359,6 +359,9 @@ def pre_process(self):
     def optimise(self):
         if self.optimised:
             return True
+        if self.triggers.value != 0:
+            # Triggers have side-effects, and can't be skipped.
```
glx22, line 363: Oh I totally missed that. I though about all (I think) non-skip cases for non random, but didn't check random.

glx22 (approved):
Seems to be the right thing to do.

PR #181, opened 2021-01-06
Fix #180: No proper error message was given, if an unreferenced Strin…

### LINKED ISSUE #180: NML erroring out when compiling
NMLC under Win 10 on 64bit machine errors out with the below text:

"C:\NML>nmlc CSBusSetN4.5.5.nml
 nmlc ERROR: nmlc: An internal error has occurred:
nmlc-version: 0.5.0
Error:    (IndexError) "pop from empty list".
Command:  ['nmlc', 'CSBusSetN4.5.5.nml']
Location: File "nml\actions\action4.py", line 170, in get_string_action4s"

I have tried multiple versions, all do the same.

Link to NML file and lang files for review if needed:
https://www.dropbox.com/sh/r3coo1l9zw9wxn2/AAAXmVNYmRtnzJiZ2cEj7REua?dl=0

glx22:
…g was unable to allocate an id

### DIFF
--- a/nml/actions/action4.py
+++ b/nml/actions/action4.py
@@ -177,8 +177,16 @@ def get_string_action4s(feature, string_range, string, id=None):
             if (feature, string) in used_strings[string_range]:
                 id_val = used_strings[string_range][(feature, string)]
             else:
-                id_val = string_ranges[string_range]["ids"].pop()
-                used_strings[string_range][(feature, string)] = id_val
+                try:
+                    id_val = string_ranges[string_range]["ids"].pop()
+                    used_strings[string_range][(feature, string)] = id_val
+                except IndexError:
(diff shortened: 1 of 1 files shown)


FLHerne:
The error message still isn't very user-friendly, it doesn't explain that the failure is due to hitting a limit nor what that limit is.

Something like the one for items would be better: "Unable to allocate ID for string, no more free IDs available (maximum is {:d})" ?

The regression-test failure is odd and seems unrelated to this change?

glx22:
Regression failure is caused by newer version of pillow.
And I agree, I'll update the message.

LordAro (approved):
LGTM

LordAro (approved):

PR #182, opened 2021-01-06
Fix: regression failure with Pillow 8.1.0

glx22:
A [change](https://github.com/python-pillow/Pillow/commit/2f409261eb1228e166868f8f0b5da5cda52e55bf) in Pillow 8.1.0 (release 4 days ago) triggers `OSError: buffer overrun when reading image file` for opengfx_generic_trams1.pcx (371x150) and `OSError: image file is truncated (0 bytes not processed)` for opengfx_trains_start.pcx (409x210)

I don't really know if it's a bug in Pillow, but I fixed the issue by croping first file to 370x150 and second file to 408x210.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
     ext_modules=[Extension("nml_lz77", ["nml/_lz77.c"], optional=True)],
     python_requires=">=3.5",
     install_requires=[
-        "Pillow>=3.4",
+        "Pillow>=3.4,<8.1.0",
         "ply",
     ],
     cmdclass={"build_py": NMLBuildPy},


LordAro:
I'm not sure I like changing the tests to fix a bug :)
At the very least, shouldn't nml handle files of "broken" sizes better?

glx22:
I'm not sure the sizes are "broken" as it worked perfectly before Pillow update.

LordAro:
No, as far as NML is concerned they are not. But the images (assuming they're not actually invalid) should not have to be changed to work around Pillow bugs/changed behaviour

And also, NML should handle such "broken" images better than outputting a python stacktrace

glx22:
I think the best for now is to not use Pillow >=8.1.0
Then the best move will be to try to write a minimal test case and submit a bug report to Pillow.

glx22:
Opened an [issue on pillow](https://github.com/python-pillow/Pillow/issues/5203)

glx22:
Got a reply. The failing PCX are indeed malformed (according to spec linked in the issue comment). But they opened a PR that may allow loading, as some image editors (like gimp) can create malformed PCX.
So still waiting.

LordAro (approved):

LordAro:
Given that it's out of spec, it's probably fine to change the images, but fixing the CI is more important right now, imo

matthijskooijman:
This fix is a bit problematic for the Debian package: We don't have the luxury of using an older pillow version, so this bug now breaks building in unstable, which is a release-critical bug.

Any plans to also fix the PCX images? Would that be a matter of re-saving them with a program that saves them properly?

glx22:
As said in PR on pillow, 
```
Unfortunately, both Gimp and ImageMagick both write images with odd strides
```
So the solution would be my first idea, ie resize the images.

FLHerne:
An issue is that lots of grfs are made using GIMP, so this issue is likely to hurt actual users. The test images are typical of grf source files in the wild and indeed taken from real projects.

Wormnest:
Can you open an issue on GIMP's issue tracker linking this then I'll try to look at what we can do on our end.

glx22:
In the issue I opened on pillow there's a link to [pcx spec](https://web.archive.org/web/20100206055706/http://www.qzx.com/pc-gpe/pcx.txt), and it seems many tools don't follow `BytesPerLine` description.

Wormnest:
Actually this was already fixed in GIMP 2 years ago, so if you have a recent enough GIMP it should not save with odd values of `BytesPerLine` anymore. Meaning you can just re export the images, resizing is not necessary.

See: [https://gitlab.gnome.org/GNOME/gimp/-/issues/2160](https://gitlab.gnome.org/GNOME/gimp/-/issues/2160)

glx22:
Indeed re exporting files with GIMP seem to be enough.

PR #183, opened 2021-01-16
Add support for recent NewGRF additions

frosch123:
Adds support for some recent NewGRF additions, those that I know what they do.

I am a bit worried about adding such a generic name like "inflation" to the global namespace. But I have no better idea either.

### DIFF
--- a/nml/actions/action2var_variables.py
+++ b/nml/actions/action2var_variables.py
@@ -563,7 +563,7 @@ def industry_town_count(name, args, pos, info):
     return (args[0], extra_params)
 
 def industry_cargotype(name, args, pos, info):
-    return (expression.functioncall.builtin_resolve_typelabel(name, args, pos), [])
+    return (expression.functioncall.builtin_resolve_typelabel(name, args, pos, table_name="cargotype"), [])
 
 varact2vars60x_industries = {
     'nearby_tile_industry_tile_id' : {'var': 0x60, 'start':  0, 'size': 16, 'param_function': unsigned_tile_offset},
--- a/nml/actions/action2var_variables.py
+++ b/nml/actions/action2var_variables.py
@@ -148,6 +148,7 @@ varact2vars_vehicles = {
     'vehicle_is_offered'               : {'var': 0x48, 'start':  2, 'size':  1},
     'build_year'                       : {'var': 0x49, 'start':  0, 'size': 32},
     'vehicle_is_potentially_powered'   : {'var': 0x4A, 'start':  8, 'size':  1},
+    'tile_has_catenary'                : {'var': 0x4A, 'start':  9, 'size':  1},
     'date_of_last_service'             : {'var': 0x4B, 'start':  0, 'size': 32},
     'position_in_articulated_veh'          : {'var': 0x4D, 'start':  0, 'size':  8},
     'position_in_articulated_veh_from_end' : {'var': 0x4D, 'start':  8, 'size':  8},
--- a/nml/actions/action2var_variables.py
+++ b/nml/actions/action2var_variables.py
@@ -178,6 +178,7 @@ varact2vars_vehicles = {
 # Vehicle-type-specific variables
 
 varact2vars_trains = {
+    **varact2vars_vehicles,
     #0x4786 / 0x10000 is an approximation of 3.5790976, the conversion factor
     #for train speed
     'max_speed'           : {'var': 0x98, 'start': 0, 'size': 16, 'value_function': value_mul_div(0x4786, 0x10000)},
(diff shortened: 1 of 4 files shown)


frosch123:
@FLHerne : I think we discussed builtin_typelabel_available before. I had to add an optional parameter to make it useable for rail/road/tram-types.

nml/expression/functioncall.py
```diff
@@ -457,25 +457,22 @@ def builtin_str2number(name, args, pos):
...
 @builtins("cargotype", "railtype", "roadtype", "tramtype")
-def builtin_resolve_typelabel(name, args, pos):
+def builtin_resolve_typelabel(name, args, pos, table_name=None):
     """
     {cargo,rail,road,tram}type(label) builtin functions.
 
     Also used from some Action2Var variables to resolve cargo labels.
     """
     tracktype_funcs = {
+        "cargotype": global_constants.cargo_numbers,
```
FLHerne, line 467: Using `cargotype` here will change error messages "...in your cargo table." to "...in your cargotype table."

Given that the syntax is `cargotable {}` vs `railtypetable {}` (etc.) this might not be a desired change.

nml/actions/action2var_variables.py
```diff
@@ -231,6 +231,15 @@ def signed_byte_parameter(name, args, pos, info):
     ret = nmlop.AND(args[0], 0xFF, pos).reduce()
     return (ret, [])
 
+def vehicle_railtype(name, args, pos, info):
+    return (expression.functioncall.builtin_resolve_typelabel(name, args, pos, table_name="railtype"), [])
```
FLHerne, line 235: This commit needs to come after "Codechange: make builtin_resolve_typelabel usable for other action2 variables".

nml/actions/action2var_variables.py
```diff
@@ -148,6 +148,7 @@ def unsigned_tile_offset(name, args, pos, info):
     'vehicle_is_offered'               : {'var': 0x48, 'start':  2, 'size':  1},
     'build_year'                       : {'var': 0x49, 'start':  0, 'size': 32},
     'vehicle_is_potentially_powered'   : {'var': 0x4A, 'start':  8, 'size':  1},
+    'tile_has_catenary'                : {'var': 0x4A, 'start':  9, 'size':  1},
```
FLHerne, line 151: Would be nice if the commit message linked to the relevant OTTD change:

https://github.com/OpenTTD/OpenTTD/commit/f1bc4d8407613795c1b0c1fbe3b1adf39479913d

nml/actions/action2var_variables.py
```diff
@@ -240,6 +249,26 @@ def signed_byte_parameter(name, args, pos, info):
     'other_veh_z_offset'  : {'var': 0x62, 'start': 24, 'size': 8, 'param_function':signed_byte_parameter, 'value_function':value_sign_extend},
 }
 
+varact2vars60x_trains = {
```
FLHerne, line 252: Again, would be helpful to link to the OTTD change:
https://github.com/OpenTTD/OpenTTD/commit/868d84bbfc8d2c1471ef713854dd05ebc872615a

nml/actions/action2var_variables.py
```diff
@@ -240,6 +249,26 @@ def signed_byte_parameter(name, args, pos, info):
...
+    'tile_knows_rail_type'    : {'var': 0x63, 'start':  0, 'size': 1, 'param_function':vehicle_railtype},
```
FLHerne, line 253: What does it mean for a tile to "know" a railtype? Looking at the OTTD code this seems to be a global 
check, in which case it duplicates `railtype_available()`

[obviously ditto for road/tram]

FLHerne, line 253: All existing properties/variables use `railtype` rather than `rail_type`.

nml/global_constants.py
```diff
@@ -1224,6 +1224,7 @@ def config_flag(name, info, pos):
     "dynamic_engines"         : 0x78,
     "variable_runningcosts"   : 0x7E,
     "256_persistent_registers": 0x80,
```
FLHerne, line 1226: Commit message again (this time https://github.com/OpenTTD/OpenTTD/commit/1478fa93b344fc25000725dd143109e4def24f0b)

nml/actions/action2var_variables.py
```diff
@@ -240,6 +249,26 @@ def signed_byte_parameter(name, args, pos, info):
...
+    'tile_supports_rail_type' : {'var': 0x63, 'start':  1, 'size': 1, 'param_function':vehicle_railtype},
+    'tile_powers_rail_type'   : {'var': 0x63, 'start':  2, 'size': 1, 'param_function':vehicle_railtype},
+    'tile_is_rail_type'       : {'var': 0x63, 'start':  3, 'size': 1, 'param_function':vehicle_railtype},
+}
+varact2vars60x_trains.update(varact2vars60x_vehicles)
```
FLHerne, line 258: Can we please use `{**varact2vars60x_vehicles, ...}` in new code rather than using update()?

(/me adds to list of potential cleanups...)

FLHerne (changes requested):

nml/actions/action2var_variables.py
frosch123: I link the OpenTTD PRs now. We also do that in the spec status tracker.

nml/actions/action2var_variables.py
frosch123: You are right. I removed the variable, and added a comment why it is skipped.

frosch123: "all" is not true. But I copied from the minority :)

FLHerne (approved):

PR #185, opened 2021-01-21
Fix #184: Share townname bits when possible

### LINKED ISSUE #184: NewGRF Town Names no longer compile after unknown changes
I am unable to compile FicTownNames Reloaded, and various other town name NewGRFs have been tested to also fail to compile under similar circumstances under nmlc 5.3...

`ERROR: "src/FicTownNames_R.nml", line 28: Not enough random bits for the town name generation (40 needed, 32 available)`

...but compile with no issues in nmlc 4.5, therefore I seem to have discovered an undocumented bug. If it's not a bug, and the bits are not meant to be used more than once, then I'm afraid you have made every sufficiently large town name GRF obsolete. Hopefully there's an easy fix for this?

I can provide my source code if needed, but it's a universal issue affecting all GRFs.

glx22:
Since #117, townname bits were not allowed to overlap. But that was extreme, as in many cases overlapping is possible.

I removed ActionF.free_bit attribute and replaced it with a deeper analyse of Action F dependencies, with updating of already prepared Action F if required.

I also added more cases to the regression test.


LordAro (dismissed):
Fine by me, perhaps @Gadg8eer should test it?

glx22:
I tested with his source, NFO output is the same as before #117.

LordAro (approved):

PR #187, opened 2021-01-30
Fix/silence warning from flake8-bugbear.

frosch123:
None of the fixes are critical. But the noise in "make flake" annoys me :)

### DIFF
--- a/nml/actions/action2layout.py
+++ b/nml/actions/action2layout.py
@@ -67,7 +67,7 @@ class Action2Layout(action2.Action2):
         file.newline()
         if len(self.sprite_list) == 0:
             file.print_dwordx(0)  # sprite number 0 == no sprite
-            for i in range(0, 5):
+            for _ in range(0, 5):
                 file.print_byte(0)  # empty bounding box. Note that number of zeros is 5, not 6
         else:
             for sprite in self.sprite_list:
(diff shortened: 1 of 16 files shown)


glx22 (approved):
LGTM

nml/actions/action2layout.py
```diff
@@ -67,7 +67,7 @@ def write(self, file):
         file.newline()
         if len(self.sprite_list) == 0:
             file.print_dwordx(0)  # sprite number 0 == no sprite
-            for i in range(0, 5):
+            for _ in range(0, 5):
```
FLHerne, line 70: I'm not a fan of this loop. What do you think of `file.print_data(b"\0" * 5)`, or just `file.print_data(b"\0\0\0\0\0")`?

FLHerne (commented):
See one comment.

I considered nitpicking about some of the AssertionErrors where semantically ValueError might make sense, but meh.

nml/actions/action2layout.py
glx22: It's a dword so each value is written \dx00000000 (in NFO) or 00 00 00 00 (in GRF)

FLHerne: Good point. Ignore this idea then.

PR #188, opened 2021-02-06
Fix broken regression PCX and enable pillow 8.1.0 with an improved error message

glx22:
As pointed in #182, GIMP now creates valid PCX. So I think it's OK to fix our broken files.

That allows to use pillow 8.1.0 as regression now pass.

As broken PCX may still be in the wild, giving a nice error message will help by displaying directly which file causes failure.


nml/spriteencoder.py
```diff
@@ -301,7 +301,11 @@ def encode_sprite(self, sprite_info):
             if x < 0 or y < 0 or x + size_x > im_width or y + size_y > im_height:
                 pos = generic.build_position(sprite_info.poslist)
                 raise generic.ScriptError("Read beyond bounds of image file '{}'".format(filename_32bpp.value), pos)
-            sprite = im.crop((x, y, x + size_x, y + size_y))
+            try:
+                sprite = im.crop((x, y, x + size_x, y + size_y))
+            except OSError:
+                pos = generic.build_position(sprite_info.poslist)
+                raise generic.ImageError("Failed to crop 32bpp {} image".format(im.format), filename_8bpp.value, pos)
```
frosch123, line 308: s/filename_8bpp/filename_32bpp/

frosch123 (approved):

matthijskooijman:
Cool, thanks for this fix. A user from Debian also confirmed that these new PCX files fix the build of the Debian package as well: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=980641#15

Are there plans for a new release with these changes included, maybe? Because this changes binary files, I believe backporting these changes is not as easy as with text files, so a new release would probably be the easiest way to get this fixed in Debian (but if not, I can probably add some scripts to update the binary files anyway).

matthijskooijman:
> Are there plans for a new release with these changes included, maybe? Because this changes binary files, I believe backporting these changes is not as easy as with text files, so a new release would probably be the easiest way to get this fixed in Debian (but if not, I can probably add some scripts to update the binary files anyway).

Never mind this, I ended up adding the scripts to backport the fixed pcx files, so this is no longer an issue for Debian. Also, Debian is in softfreeze, so it's unlikely that any upstream version newer than 0.5.3 (unless it is really a consise bugfix-only release) will be used for the upcoming Debian bullseye.

PR #189, opened 2021-02-20
Change: Check that user code doesn't try to use reserved registers

FLHerne:
We use this range for internal calculations, so user code could get its
values overwritten by ours or vice versa.

Of course this won't catch dynamically-calculated addresses.

### DIFF
--- a/nml/expression/storage_op.py
+++ b/nml/expression/storage_op.py
@@ -19,10 +19,10 @@ from .base_expression import ConstantNumeric, Expression, Type
 from .parameter import parse_string_to_dword
 
 storage_op_info = {
-    "STORE_PERM": {"store": True, "perm": True, "grfid": False, "max": 0xFF},
-    "STORE_TEMP": {"store": True, "perm": False, "grfid": False, "max": 0x10F},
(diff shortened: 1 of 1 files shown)


frosch123:
I looked up the origin of "0xFF: Used for some house variables".
This refers to "house_same_class" in action2var_variables.
So this variable is also reserved by NML.

nml/expression/storage_op.py
```diff
@@ -111,6 +111,10 @@ def reduce(self, id_dicts=None, unknown_id_fatal=True):
             raise generic.ScriptError("Register to access must be an integer.", register.pos)
         if isinstance(register, ConstantNumeric) and register.value > self.info["max"]:
             raise generic.ScriptError("Maximum register for {} is {:d}".format(self.name, self.info["max"]), self.pos)
+        if isinstance(register, ConstantNumeric) and 0x80 <= register.value <= 0xFF:
```
frosch123, line 114: I think this needs a "not self.info["perm"]", otherwise it triggers for persistent storage.
Though maybe it's also smart to extend "self.info" with some "reserved" member or something.

frosch123 (approved):

PR #199, opened 2021-04-01
Add: support of square root and round functions to builtin functions

werbfred:
Added sqrt function to builtin ones as it could be usefull in other projects.

I implemented it because the SQRT approximation in 2cc_TrainsInNML.grf was giving sometimes weird results. Using math.sqrt is much more accurate.

### DIFF
--- a/nml/expression/functioncall.py
+++ b/nml/expression/functioncall.py
@@ -534,22 +534,33 @@ def builtin_cargoexpr(name, args, pos):
         raise AssertionError()
 
 
-@builtins("acos", "asin", "atan", "cos", "sin", "tan")
-def builtin_trigonometric(name, args, pos):
+@builtins("acos", "asin", "atan", "cos", "sin", "sqrt", "tan")
+def builtin_math(name, args, pos):
     if len(args) != 1:
         raise generic.ScriptError(name + "() must have 1 parameter", pos)
     val = args[0].reduce()
     if not isinstance(val, (ConstantNumeric, ConstantFloat)):
         raise generic.ScriptError("Parameter for " + name + "() must be a constant", pos)
-    trigonometric_func_table = {
+    math_func_table = {
         "acos": math.acos,
         "asin": math.asin,
         "atan": math.atan,
         "cos": math.cos,
         "sin": math.sin,
+        "sqrt": math.sqrt,
         "tan": math.tan,
     }
-    return ConstantFloat(trigonometric_func_table[name](val.value), val.pos)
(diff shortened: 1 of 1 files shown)


nml/expression/functioncall.py
```diff
@@ -552,6 +552,15 @@ def builtin_trigonometric(name, args, pos):
     return ConstantFloat(trigonometric_func_table[name](val.value), val.pos)
 
 
+@builtin
+def builtin_sqrt(name, args, pos):
+    if len(args) == 0:
+        raise generic.ScriptError("sqrt() requires 1 argument", pos)
+    elif len(args) != 1:
+        raise generic.ScriptError("sqrt() requires only 1 argument", pos)
+    return ConstantFloat(math.sqrt(args[0].value),pos)
```
frosch123, line 561: This probably needs a check for args[0] being a constant value, similar as for the trigonometric functions above.
```
if not isinstance(val, (ConstantNumeric, ConstantFloat)):
        raise generic.ScriptError("Parameter for " + name + "() must be a constant", pos)
```

Actually, it may be easier to just extend the mapping in builtin_trigonometric, and rename it to builtin_math or something.

werbfred:
Added **round** builtin as well as it's better than using **int**. This way floats are rounds down/up to its nearest integer.

In my use case, this allows using formulas to determine base/running costs more accurate and ensuring them to be an 1 Byte integer.

nml/expression/functioncall.py
werbfred: Updated Pull Request's commit with proposed updates

FLHerne:
You can simply add `round` to `builtin_math` also? It's exactly the same code.

Please reorder or squash the commits so you're not introducing an error and then fixing it. That should also make the commit-checker stop complaining.

werbfred:
**Round** should return a **_ConstantNumeric_** i.e. an integer value, while all other math functions as **sqrt** and existing **sin**, **cos**, **tan** return **_ConstantFloat_** s...

FLHerne:
> **Round** should return a **_ConstantNumeric_** i.e. an integer value, while all other math functions as **sqrt** and existing **sin**, **cos**, **tan** return **_ConstantFloat_** s...

Sorry, I missed that. My mistake.

frosch123 (approved):

frosch123:
Docs:
https://newgrf-specs.tt-wiki.net/index.php?title=NML%3ABuiltin_functions&type=revision&diff=4100&oldid=4039

PR #200, opened 2021-04-04
Change: Display input filename

### LINKED ISSUE #194: nmlc doesn't throw an error when GRF name missing
When you leave a GRF name out of the command, i.e. "nmlc --grf myfile.nml"  the program hangs on "Reading..." instead of throwing an error or giving the GRF a default name. If the command is entered incorrectly it should display an error because I incorrectly assumed it would work the same way as another program and wasted an embarrassing amount of time trying to fix it.

glx22:
As suggested in #194, this replaces `Reading ...` message with `Reading filename.nml ...` or `Reading standard input ...` depending on whether an input filename was given.

Closes #194.

### DIFF
--- a/nml/main.py
+++ b/nml/main.py
@@ -378,7 +378,7 @@ def nml(
     """
     generic.OnlyOnce.clear()
 
-    generic.print_progress("Reading ...")
+    generic.print_progress("Reading {} ...".format(input_filename or "standard input"))
 
     try:
         script = inputfile.read()


nml/main.py
```diff
@@ -378,7 +378,7 @@ def nml(
     """
     generic.OnlyOnce.clear()
 
-    generic.print_progress("Reading ...")
+    generic.print_progress("Reading {} ...".format("standard input" if input_filename is None else input_filename))
```
frosch123, line 381: How about
```suggestion
    generic.print_progress("Reading {} ...".format(input_filename or "standard input"))
```

frosch123 (approved):

PR #207, opened 2021-04-23
Fix #206: Apply correct mask to `town_euclidean_dist` variables

### LINKED ISSUE #206: town_euclidean_dist  - Industry variable returns wrong value
town_euclidean_dist (0x66) variable for industries returns value with town zone included like this: zzdddd, so the result value is wrong.
It should be masked with 0xFFFF to return only the distance. Otherwise the value has to be masked in nml newgrf code, but then it should be stated in nml wiki.

Check action2var_variables.py, line 608 agains newgrf_industries.cpp line 290. 
Size should be probably fixed to 16 in nml source.

glx22:
`town_euclidean_dist` (object var 46, industry var 66) was not following the same pattern as `town_manhattan_dist` (object var 45, industry var 65) while the format of these 4 variables is the same.


LordAro (approved):
Sure.

PR #212, opened 2021-04-24
Fix #209: Sign-extend value before cargo_profit callback result calculation

### LINKED ISSUE #209: Cargo callback "profit" probably broken
Cargo callback "**profit**" behaves strangely.

The problem is that when returning negative values, the cargo payment is for some values actually positive. Not sure if that is intended, but seems counter-intuitive.

Also cargo property "**price_factor**" is strange. Wiki states it is float type, but does not tells any lmit. In the NFO it is DWORD, so basically UINT32. 
Anyway, its behaviour is not linear. When its value is increased lineary, the cargo payment sometimes drops back to lower values.

glx22:
NML does apply some calculation before returning the cargo_profit callback result.
But it does it on a 15bit signed value stored in 32bit without sign-extending, so of course the result for negative is wrong.

It's fixed in this PR by applying sign-extension to the 15bit value before doing other calculation.

Of course another option would be to just remove all this strange magical stuff ;)

### DIFF
--- a/nml/actions/action3_callbacks.py
+++ b/nml/actions/action3_callbacks.py
@@ -216,6 +216,9 @@ def cargo_profit_value(value):
     # (amount * price_factor * (cb_result * 329 / 256)) / 256
     # This allows us to report a factor of 256 in the documentation, which makes a lot more sense than 199.804...
     # Not doing the division here would improve accuracy, but limits the range of the return value too much
+    # As cb_result is a 15bit signed value, we need to sign-extend it to 32bit before applying our math
+    value = nmlop.SHIFT_LEFT(value, 17)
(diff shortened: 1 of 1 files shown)


PeterN:
+1 here for removing this strange mapping. I can't see that it serves any purpose. Compatibility may be an issue, not sure how to find out how well used this callback is.

PeterN:
I put feelers out on the NewGRF forum about this callback and... https://www.tt-forums.net/viewtopic.php?f=68&t=88938

Nothing.

andythenorth:
Per https://github.com/OpenTTD/nml/pull/193#issuecomment-877127255 Tyler intends to use cargo_profit cb.

PR #216, opened 2021-05-04
Fix #214: Prevent scope change for function calls when optimising switches

### LINKED ISSUE #214: Issue with passing a variable to a function from the PARENT scope
Apologies in advance for any misuse of terminology here...

Note: I am using PR #173 to build, but I don't think that affects this

When trying to pass a variable to a function which exists in the PARENT scope of the feature I'm calling from, I could only make it work by using a dummy switch in-between (or else it throws an error, due to being the wrong scope). This isn't a big deal, except it results in nmlc trying to "optimise" the switch (since it isn't doing anything except sending the variable), which results in the variable being apparently discarded and the callback failing to work properly.

I was able to work around this by making the dummy switch do some garbage operation (in this case store a value to a temp register) to prevent the "optimising" and make it work, but it seems like this shouldn't be necessary. Example below of what I'm doing:

```
switch (FEAT_INDUSTRYTILES, PARENT, switch_mineral_mine_prod_change_2, STORE_TEMP(0, 2)) {
	switch_primary_industry_prod_change(transported_last_month_pct("MNRL"));
}
item (FEAT_INDUSTRYTILES, mineral_mine_tiles_16) {
	property {
		substitute:          87;
//		override:            ?;
		accepted_cargos:     [[MAIL, 8]];
		land_shape_flags:    bitmask(LSF_ONLY_ON_FLAT_LAND, LSF_CANNOT_LOWER_NW_EDGE, LSF_CANNOT_LOWER_NE_EDGE, LSF_CANNOT_LOWER_SW_EDGE, LSF_CANNOT_LOWER_SE_EDGE);
	}
	graphics {
		default:             switch_mineral_mine_tiles_16_decide_ground_sprite;
		random_trigger:      switch_mineral_mine_prod_change_2;
		autoslope:           CB_RESULT_NO_AUTOSLOPE;
	}
}
```

Am I just doing this completely wrong? Is there a better way to do this without the use of a dummy switch? It's a bit hard to tell since there doesn't seem to be any obvious documentation on this feature, so I'm just taking shots in the dark here.

If this is "correct", it would seem that perhaps nmlc is being a bit overzealous with it's optimizations in this case, and that function calls should be ignored somehow. But I'd appreciate any advice to get me on the right track if not 🙂

glx22:
Fixes #214

A procedure call can use variables from current scope as call parameters.
When the switch is optimised as a procedure call, a return action in caller's scope is created, and wrong variables may be accessed.

Prevent this by considering procedure calls like any complex expression, and allow the switch to be rewritten.
~~It's not needed for random switches as variables are always accessed in `SELF` scope for them.~~
Don't optimise procedure calls in random switches.

And while I was updating a regression test, I decided to include string test for #205.


LordAro (approved):
Code is fine

Andrew350:
So, I managed to frankenstein this PR together with current master and PR #173, and it appears to work, so thanks for a quick fix 🙂. One thing I noticed: The .grf works as expected and I get no errors when using the switch in this format:

```
switch (FEAT_INDUSTRYTILES, PARENT, switch_mineral_mine_prod_change_2, switch_primary_industry_prod_change(transported_last_month_pct("MNRL"))) {
	return;
}
```

However if I write it similar to the way shown in the original issue, but with the 4th parameter simply empty brackets and the call as the return value...

```
switch (FEAT_INDUSTRYTILES, PARENT, switch_mineral_mine_prod_change_2, []) {
	switch_primary_industry_prod_change(transported_last_month_pct("MNRL"));
}
```
...I don't get any syntax errors but nmlc still "optimises" the switch resulting in failure. I'm not really sure if this is even valid code, or related to this issue, but thought I should mention it.

glx22:
Does it happen with or without my fix ?
With my fix the second version is rewritten to the first version during optimisation.
Without my fix the second version is translated in a return action (same format as first version, but in caller's scope).

Andrew350:
This is with your fix applied, but also PR #173 applied as well (as I can't build without it). I'd have to build a special GRF to check your fix separately, but they don't seem to conflict here so not sure if that affects anything.

PR #220, opened 2021-06-05
Fix #209: Cargo 'profit' callback applied a unit conversion, when there was no unit.

### LINKED ISSUE #209: Cargo callback "profit" probably broken

frosch123:
This is an alternative to #212 .

The documentation for the callback and the comments in NML are very technical.
After this PR the description can be simplified to:
```
cargo_income = callback_result * cargo_amount * price_factor
```

To do something similar as the default-payment, you can use
```
callback_result = distance * time_penalty
```
with 'time_penalty' in range 0..1.

Closed #212


frosch123:
I checked all NewGRF hosted on devzone. None of them use this callback.

PeterN:
Seems like the right thing to do to me.

andythenorth:
Per https://github.com/OpenTTD/nml/pull/193#issuecomment-877127255 Tyler intends to use cargo_profit cb.

2TallTyler:
I'd like to start using the cargo_profit callback in [Industries of the Caribbean](https://github.com/2TallTyler/industries_of_the_caribbean). If I fork NML and apply this commit, will negative values still be calculated properly? (not requiring #212?)

LordAro (approved):
#YOLO

2TallTyler:
\o/

PR #222, opened 2021-06-12
Feature: Support for rail vehicle property 0x2E, curve_speed_mod

vituscze:
Depends on https://github.com/OpenTTD/OpenTTD/pull/9346


michicc:
Might be fine. Negative values are correctly translated and the scaling factor seems to fit, too.

andythenorth (approved):
Based on Michi's  comment, approving.  The only area of concern was 255 vs 256 for unit conversion (don't know which is correct).

FLHerne (approved):
From when we discussed this on IRC, I'm fairly sure 256 is correct. If it isn't, it's already wrong for several other properties with the same value range and would need fixing anyway.

PR #223, opened 2021-06-24
Add: town var 0x41 'town_index'

andythenorth:
Per the comment at https://newgrf-specs.tt-wiki.net/wiki/NML:Deprecated_syntax

Rather than relying on ``var[0x41, 0, 0xFFFF]``, this PR adds a var ``town_index``

Town var 0x41 is a word, at least for OpenTTD purposes: https://newgrf-specs.tt-wiki.net/wiki/VariationalAction2/Towns

Tested with some FIRS code which is not helpful to other people as it has a bunch of unrelated changes, but the var works as expected.


LordAro (approved):
Sure.

andythenorth:
Added docs https://newgrf-specs.tt-wiki.net/wiki/NML:Towns

PR #224, opened 2021-07-10
Fix: Reject empty arrays of expressions

glx22:
@andythenorth found a bug ;)
```
switch(FEAT_INDUSTRIES, SELF, test_empty_expression_switch,
    // NML was throwing a not-descriptive parse error with an empty expression
    []) {
    return;
}
```
triggers "IndexError: pop from empty list" at https://github.com/OpenTTD/nml/blob/840613eb393eface5026dd15e491051176f6dc3c/nml/actions/action2var.py#L639.

But it really should have been rejected way earlier.

### DIFF
--- a/nml/actions/action2var.py
+++ b/nml/actions/action2var.py
@@ -631,6 +631,8 @@ class Varaction2Parser:
 
     def parse_expr(self, expr):
         if isinstance(expr, expression.Array):
+            if len(expr.values) == 0:
+                raise generic.ScriptError("An array of expressions cannot be empty", expr.pos)
             for expr2 in expr.values:
                 self.parse(expr2)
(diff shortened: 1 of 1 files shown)


andythenorth:
Confirmed works as expected with FIRS test case.

Empty array now produces:

` nmlc ERROR: "generated/firs.nml", line 185561: Switch-block last parameter 'expression' cannot be an empty array.`

andythenorth (dismissed):

andythenorth:
Tested the revised version, works as expected with FIRS test case.

LordAro:
Can we add a regression test?

glx22:
Quite impossible to add regression test (our regression system doesn't expect failures).

andythenorth:
Yolo?

PR #227, opened 2021-07-17
Fix f2cab72: Called procedures may receive wrong values in parameters

glx22:
When temporary intermediate variables are used, register number was passed as call parameter instead of the register content.

It's quite obvious in NFO diffs (`\2r 1A` vs `\2r 7D`), but I totally missed it before.


andythenorth:
Fixes this FIRS test case as expected.  https://gist.github.com/andythenorth/df51f3b8fa7bce212a87604702fb836c

frosch123 (approved):

PR #230, opened 2021-08-07
Add: Options to disable optimisation and/or deprecation warnings

glx22:
This allows to silence some annoying warnings.

### DIFF
--- a/nml/actions/action0.py
+++ b/nml/actions/action0.py
@@ -476,10 +476,12 @@ def get_property_info_list(feature, name):
     for prop_info in prop_info_list:
         if "replaced_by" in prop_info:
             generic.print_warning(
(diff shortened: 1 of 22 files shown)


LordAro:
:+1: Though can't help but feel like having the type parameter last is a bit weird, and easily missed. But obviously putting it first is infeasible at this point (I imagine there are a lot of warning messages) Perhaps a few helper functions?

andythenorth:
--no-optimisation-warning works for me with FIRS

glx22:
> 
> 
> 👍 Though can't help but feel like having the type parameter last is a bit weird, and easily missed. But obviously putting it first is infeasible at this point (I imagine there are a lot of warning messages) Perhaps a few helper functions?

Well, there are 40 references of `print_warning`, so it should not be too hard to add type as first parameter.

LordAro (approved):

PR #231, opened 2021-08-14
Change: add string pluralisation examples to example industry

andythenorth:
May be useful, may not.  Documentation and test cases for string pluralisation is generally sparse at time of writing.


LordAro:
Some regression reference file updates needed ;)

andythenorth:
Probably close this as TMWFTLB and copy the examples to wiki page instead.

PR #236, opened 2021-09-01
Add: warn if a required property is not set for objects (feature 0F)

### LINKED ISSUE #191: Does not error when defining an object without a size
NML should probably error in these cases.

See OpenTTD/OpenTTD#8276

glx22:
Other features may use this, but I only checked ttdpatch source for objects.

Closes #191.


FLHerne:
I think this should be an error and not a warning, since there's no question it's incorrect and will fail.

nml/actions/action0properties.py
```diff
@@ -1274,9 +1276,9 @@ def object_size(value):
     "class":                  {"size": 4, "num": 0x08, "first": None, "string_literal": 4},
     # strings might be according to specs be either 0xD0 or 0xD4
     "classname":              {"size": 2, "num": 0x09, "string": 0xD0},
-    "name":                   {"size": 2, "num": 0x0A, "string": 0xD0},
-    "climates_available":     {"size": 1, "num": 0x0B},
-    "size":                   {"custom_function": object_size},  # = prop 0C
+    "name":                   {"size": 2, "num": 0x0A, "string": 0xD0, "required": None},
+    "climates_available":     {"size": 1, "num": 0x0B, "required": None},
+    "size":                   {"custom_function": object_size, "required": None},  # = prop 0C
```
LordAro, line 1281: I'd use True here to be clearer, even though the value doesn't actually matter

planetmaker:
I'd second the comments by FLHerne and LordAro. Both make it clearer and better readable.

glx22:
Maybe I should also consider "first" as "required.

And I guess nmlc should be smart enough to move a "first" property on first place in the list (and still warn of course).

LordAro (approved):

PR #245, opened 2021-11-14
Change: Export keyword JSON for VSCode

spnda:
For the VSCode extension I provide suggestions, however, for this I require a list of all keywords. This was previously hardcoded but I think it's best if this is also generated through the generator. I periodically let a [Actions script](https://github.com/spnda/VSC-NML/blob/master/.github/workflows/syntax_update.yml) generate these JSONs to keep the extension updated. Would be nice if this could get merged asap :)


glx22 (approved):

glx22 (approved):

spnda:
thanks for finally addressing this PR :D

PR #247, opened 2022-01-15
Add: Translate ternary operator into varact2 if needed

glx22:
This is a follow-up of #243.
Instead of issuing a warning, I think it's better to translate the ternary into a procedure and call it.
I also noticed the check for potential side effects was incorrect (fixed now).


andythenorth:
When this is done, might want to update the note about ternary at https://newgrf-specs.tt-wiki.net/wiki/NML:Elementary_values

glx22:
Maybe I should just always translate the ternary into procedure, unless it's the `guard ? constant : constant` one.

JGRennison:
Turning it into a procedure would typically cost 3 extra groups (unless these can be de-duplicated somehow). There could be a risk of running into group ID limit issues if this is done too often.
There's a trade-off between avoiding redundant operations and the extra memory and associated runtime cost of extra groups.
Probably some sort of heuristic which is neither never nor always is worth pursuing?

glx22:
Using `switch(FEAT_INDUSTRIES, SELF, ternary_optimisation, proc_call_optimisation() ? 1 : proc_call_optimisation()) { return 5; }` from this PR as reference, in master the NFO is (including side effect because `proc_call_optimisation()` is not read only)
```
 // Name: ternary_optimisation
5 * 91 02 0A FF 89
7E FF 20 \dxFFFFFFFF 	// proc_call_optimisation
\2u< 1A 20 \dx00000001
\2sto 1A 20 \dx00000080 	// guard
\2^ 1A 20 \dx00000001
\2sto 1A 20 \dx00000081 	// !guard
\2r 7E FF 20 \dxFFFFFFFF 	// proc_call_optimisation
\2* 7D 81 20 \dxFFFFFFFF
\2sto 1A 20 \dx00000082
\2r 7D 80 20 \dxFFFFFFFF
\2+ 7D 82 00 \dxFFFFFFFF
\b1
\wx8000 \dx00000001 \dx00000000 	// Bogus range to avoid nvar == 0
\wx8005 // default: return 5;
```
while this PR gives
```
// Name: @ternary_action_0
5 * 24 02 0A FF 89
7E FF 00 \dxFFFFFFFF 	// proc_call_optimisation
\b1
\wx00FF \dx00000000 \dx00000000 	// proc_call_optimisation;
\wx8001 // return 1;

// Name: ternary_optimisation
6 * 24 02 0A FF 89
7E FF 00 \dxFFFFFFFF 	// @ternary_action_0
\b1
\wx8000 \dx00000001 \dx00000000 	// Bogus range to avoid nvar == 0
\wx8005 // default: return 5;
```
basically half the size (from 91 to 2*24) for a not so complex ternary.

The `switch(FEAT_INDUSTRIES, SELF, ternary_optimisation, proc_call_optimisation() ? 1 : 0) { return 5; }` from master (a `guard ? constant : constant` one) outputs
```
// Name: ternary_optimisation
5 * 45 02 0A FF 89
7E FF 20 \dxFFFFFFFF 	// proc_call_optimisation
\2u< 1A 20 \dx00000001
\2* 1A 20 \dx00000001 	// expr1 - expr2
\2+ 1A 00 \dx00000000
\b1
\wx8000 \dx00000001 \dx00000000 	// Bogus range to avoid nvar == 0
\wx8005 // default: return 5;
```

Only one extra ID might be needed if both branches need to be extracted. But that of course needs to be tested on real world samples.

nml/actions/action2var.py
```diff
@@ -819,6 +823,46 @@ def parse_minmax(value, unit_str, action_list, act6, offset):
 return_action_id = 0
 
 
+def create_ternary_action(guard, expr_true, expr_false, action_list, feature, var_scope):
+    act6 = action6.Action6()
+
+    var_range = 0x8A if get_scope(feature, 0x8A) == var_scope else 0x89
```
frosch123, line 829: `var_scope` will be the same for parent and self scope in case of vehicles, so converting `var_scope` back to `type_byte`/`var_range` won't work. `type_byte` needs to be passed separately from somewhere.

frosch123 (approved):

PR #250, opened 2022-03-13
Add: Map seed in patch variable 0x17.

michicc:
Support for [OpenTTD/OpenTTD#9834](https://github.com/OpenTTD/OpenTTD/pull/9834)


andythenorth:
Did a crude test
- nmlc compiles
- nml file using `map_seed` compiles using patched nmlc 
- `map_seed` returns a non-zero value (I didn't confirm the map seed returns accurately as I couldn't think of a quick test for that)

Suggest merging this. Easy fix later if we missed something.

FLHerne (approved):

PR #254, opened 2022-06-10
Change: skip realsprite palette validation step iff output is nfo AND…

andythenorth:
… a palette is specified as run-time parameter

Saves ~10% of compile time for me by skipping redundant palette check when using nfo output (grfcodec will validate the palettes later).

### DIFF
--- a/nml/main.py
+++ b/nml/main.py
@@ -67,9 +67,14 @@ def parse_cli(argv):
     @return: Options, and input filename (if not supplied, use C{sys.stdin}).
     @rtype:  C{tuple} (C{Object}, C{str} or C{None})
     """
-    usage = "Usage: %prog [options] <filename>\n" "Where <filename> is the nml file to parse"
+    usage = (
+        "Usage: %prog [options] <filename>\n"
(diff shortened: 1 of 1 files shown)


andythenorth:
Diff looks wrong currently with respect to black formatting.

Eddi-z:
isn't this logic backwards? if i give a palette as parameter, i'd expect that the palette gets checked, otherwise i don't care?

also, does grfcodec do the same kind of checks on the palette as nmlc?

andythenorth:
> isn't this logic backwards? if i give a palette as parameter, i'd expect that the palette gets checked, otherwise i don't care?

Fail safe.
Vanishingly few nml users will ever need or want to disable that check.

> also, does grfcodec do the same kind of checks on the palette as nmlc?

No.  It checks for invalid white pixels and invalid colour profiles.  It does not validate, e.g. animated pixels etc.

andythenorth:
Superceded by #322 - this route involves inferring that we want to skip validation if we're outputting .nfo, which is unwise.

PR #255, opened 2022-06-11
Fix #253: Use non-ternary varaction2 sequence for abs() builtin function

### LINKED ISSUE #253: abs builtin function generates excessively verbose varaction2 output
At present the abs builtin function is implemented as:
`abs(x) --> (x < 0) ? (0 - x) : x`

If x happens to be some long or expensive calculation/variable access, it gets pasted into the output 3 times and executed at run-time 3 times.
The documentation for the abs function does not hint that it is so expensive and without inspecting the output GRF authors would have no reason to suspect that it is.

Suggestions:
1. STO_TMP and variable 7D should be used such that only one copy is required of the input

or

2. A shorter non-ternary sequence should be used such as: STO_TMP, CMP with constant 0, SUB constant 1, MUL var 7D

JGRennison:
This avoids the function argument being replicated in the output and being executed at run-time three times.

### DIFF
--- a/nml/actions/action2var.py
+++ b/nml/actions/action2var.py
@@ -440,6 +440,24 @@ class Varaction2Parser:
             expr2 = nmlop.MUL(expr.expr2, VarAction2LoadTempVar(inverted_guard_var))
             return nmlop.ADD(expr1, expr2)
 
+    def preprocess_absop(self, expr):
+        assert isinstance(expr, expression.AbsOp)
+        self.parse(expr.expr.reduce())
+        input_var = VarAction2StoreTempVar()
+        input_var.comment = "input"
+        self.var_list.append(nmlop.STO_TMP)
+        self.var_list.append(input_var)
+        self.var_list.append(nmlop.VACT2_CMP)
+        var = VarAction2Var(0x1A, 0, 0)
+        self.var_list.append(var)
+        self.var_list.append(nmlop.SUB)
+        var2 = VarAction2Var(0x1A, 0, 1)
+        self.var_list.append(var2)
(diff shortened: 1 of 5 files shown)


glx22 (approved):
I did a quick test by changing https://github.com/OpenTTD/nml/blob/master/regression/019_switch.nml#L10 from `return 0` to `return abs(current_year)`

Before:
```
// Name: @return_action_0
4 * 107 02 0A FF 89
24 20 \dxFFFFFFFF
\2cmp 1A 20 \dx00000000
\2< 1A 20 \dx00000001
\2^ 1A 20 \dx00000001
\2sto 1A 20 \dx00000080 	// guard
\2^ 1A 20 \dx00000001
\2sto 1A 20 \dx00000081 	// !guard
\2r 7D 81 20 \dxFFFFFFFF
\2* 24 20 \dxFFFFFFFF
\2sto 1A 20 \dx00000082
\2r 1A 20 \dx00000000
\2- 24 20 \dxFFFFFFFF
\2* 7D 80 20 \dxFFFFFFFF
\2+ 7D 82 00 \dxFFFFFFFF
\b0
\wx8000 // Return computed value
```
After:
```
// Name: @return_action_0
4 * 42 02 0A FF 89
24 20 \dxFFFFFFFF
\2sto 1A 20 \dx00000080 	// input
\2cmp 1A 20 \dx00000000
\2- 1A 20 \dx00000001
\2* 7D 80 00 \dxFFFFFFFF
\b0
\wx8000 // Return computed value
```
And this was a very simple expression, I can see how it's even better for size and execution with a complex expression.

PR #263, opened 2022-09-01
Change: use an array of cargoes for cargo_random_triggers

glx22:
It's simpler for users to list the cargoes, than to write the bitmask manually.

After looking at OpenTTD source for station prop12, I think this implementation is correct.


frosch123 (approved):
Let's hope "Parameters of 'bitmask' cannot be greater than 31" is a good-enough error message.

PR #264, opened 2022-09-03
Change: disabled_platforms/length is now bitmask(1-8)>>1

glx22:
Writing `bitmask(2)` to disable platform/length 3 is not very intuitive. Instead we expect the user will write `bitmask(3)` and we shift the result.


frosch123 (approved):
Maybe drop the ">> 1" from the commit message.
The user now enters "bitmask(1-8)".

PR #266, opened 2022-09-05
Fix: Station animation triggers every 250 ticks, not 256.

frosch123:
The synchronized-tile-trigger is called from the station-acceptance-update cycle, not from the tile loop.


glx22 (approved):
Looks like I copy pasted too much

PR #268, opened 2022-09-10
Change: allow any expression for station (purchase_)prepare_layout

glx22:
Only simple expression, or procedure call, were allowed.
Now an array of expression can be used too.
For `purchase_prepare_layout` the expression is simply inlined into the purchase action2.
For `prepare_layout` a procedure is generated if it's not already a procedure call, this is done to reduce code duplication as `prepare_layout` is added to all cargo action2.

I decided to disable LOAD_TEMP() optimisation in spritelayout for stations, so usage of STORE_TEMP() in default, custom or foundations will not conflict.


nml/actions/action2layout.py
```diff
@@ -208,7 +208,8 @@ def get_all_registers(self):
 
     def create_register(self, name, value):
         if (
-            isinstance(value, expression.StorageOp)
+            self.feature != 0x04
```
frosch123, line 211: Maybe a comment would be helpful here on why stations are excluded.

frosch123 (dismissed):

nml/actions/action2layout.py
glx22: Like 
```suggestion
            self.feature != 0x04 # Stations may use STORE_TEMP() after the layout registers init.
```
?

frosch123 (dismissed):

frosch123 (approved):

PR #271, opened 2022-12-03
More issues in release workflow

glx22:
Adding `--verbose` to pip command helped to find the manylinux issue, keeping it will not hurt :)
Manylinux issue was
 ```
  fatal: detected dubious ownership in repository at '/github/workspace'
  To add an exception for this directory, call:

  	git config --global --add safe.directory /github/workspace
  fatal: detected dubious ownership in repository at '/github/workspace'
  To add an exception for this directory, call:

  	git config --global --add safe.directory /github/workspace
  Git checkout found but cannot determine its version. Error:  Command '['git', '-C', '/github/workspace', 'diff-index', 'HEAD']' returned non-zero exit status 128.
```
So no surprised it decided to use 0.0.0 as version.

While fixing stuff `set-output` is deprecated.
Will also need to replace `actions/upload-release-asset` as it's not maintained and uses deprecated node.js (but no hurry with our not so frequent releases)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,7 +84,7 @@ jobs:
         twine upload --username __token__ dist/*.whl dist/*.tar.gz
         version=`python setup.py --version`
         archive=`echo "nml-$version.tar.gz"`
-        echo "::set-output name=archive::$archive"
+        echo "archive=$archive" >> $GITHUB_OUTPUT
       id: nml_source_archive
 
     - name: Publish source tarball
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/release.yml [hunk 1]
```diff
@@ -35,7 +35,8 @@ jobs:
       uses: RalfG/python-wheels-manylinux-build@v0.6.0-manylinux2014_x86_64
       with:
         python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38'
-        build-requirements: '-e .'  # pip args
+        build-requirements: '-e . --verbose'  # pip args
+        pre-build-command: 'git config --global --add safe.directory /github/workspace'
```
LordAro, line 39: Should probably use CI_PROJECT_DIR (or whatever the variable is)

LordAro (approved):

PR #273, opened 2022-12-13
Add: 'tunnels' callback for road- and tramtypes

glx22:
This addition reflects OpenTTD/OpenTTD#10233

Usage is the same as railtype `tunnels`.


FLHerne (approved):
I don't really like the vague `tunnels` name, but it matches the one in railtypes like most of the other roadtype names. I'd have suggested `tunnel_overlay` but that's already used in railtypes for portal overlays so would be much more confusing.

andythenorth:
Added to nml spec https://newgrf-specs.tt-wiki.net/wiki/NML:Roadtypes#Roadtype_callbacks

Somebody had already add it, but commented out - thanks, saved time :)

PR #276, opened 2023-01-30
Feature: Engine name callback

michicc:
## Description

See [OpenTTD/OpenTTD#10399](https://github.com/OpenTTD/OpenTTD/pull/10399).

Add basic support for the engine name callback and the extended callback flag properties.

## Limitations

To make full use of the name callback in view of the limited D0xx text are that is available as a callback results, the OpenTTD PR also increases the D8xx text are up to FFFF (using up some TTDpatch specific IDs). Inclusion via the text stack and codes 80/81 can use all IDs and thus also D8xx.
So far, NML can not make use of this though, as all strings that are referred from Action 2s seem to be strictly allocated into the D0xx area. Reworking this is above my pay grade and thus not part of this PR.


nml/actions/action0.py
```diff
@@ -1119,11 +1119,9 @@ def get_callback_flags_actions(feature, id, flags):
     if not isinstance(prop_info_list, list):
         prop_info_list = [prop_info_list]
     for prop_info in prop_info_list:
-        act0.prop_list.append(
-            Action0Property(
-                prop_info["num"], parse_property_value(prop_info, expression.ConstantNumeric(flags)), prop_info["size"]
-            )
-        )
+        parsed_value = parse_property_value(prop_info, expression.ConstantNumeric(flags))
+        if "skip" not in prop_info or parsed_value[0].value != 0:
```
glx22, line 1123: I don't think the `skip` stuff is needed.
Not outputting flag prop when value is 0 should be fine with all features, and would be a small improvement in grf size.

michicc: So you mean just never output a 0 cb prop, no matter if old or new prop?

glx22: Yes, I think it should be fine as the prop value is or'ed or sb'ed on read (at least in openttd)

glx22 (approved):

andythenorth:
nml spec updated https://newgrf-specs.tt-wiki.net/wiki/NML:Vehicles#Vehicle_callbacks

PR #277, opened 2023-02-17
Fix: By default there is no ELRL railtype.

frosch123:
To define ELRL trains, you can either
* define a railtype table containing ELRL, or
* use track_type=RAIL with engine_class=ENGINE_CLASS_ELECTRIC.

This was already reflected in the documentation.

In no case does ELRL match MONO.


Brickblock1:
You should probably change that 4 to a 3 while you are at it. : )

glx22 (approved):

PR #281, opened 2023-04-01
Optimise variational action2 trivial cases

JGRennison:
Drop useless binary operator operations with constant 0 for:
* shift left/right
* or
* xor
* rotate

Optimise trivial right shifts of variables by a constant value. Adjust the shift right and mask of a variable where possible, in the same way as is currently done for left shifts of inherently right shifted variables.

The practical upshot of this is that extremely common patterns like: `getbits(extra_callback_info2, 0, 8)` now result in the obvious representation:
* ADD/RST variable 18, shift: 0, mask: FF

Instead of currently:
* ADD/RST variable 18, shift: 0, mask: FFFFFFFF
* SHR variable 1A, shift: 0, mask: 0
* AND variable 1A, shift: 0, mask: FF

Using one instead of three ops allows direct references in more cases instead of needing to spill to a temporary register and use RST.


glx22:
Regression fails, but the output clearly shows the optimisation.

FLHerne:
Nice improvement.

This looks correct to me - just need to update regression/expected/* with the new results.

glx22 (approved):

PR #283, opened 2023-04-16
Add: STAT_ALL_TILES constant for use instead of 0xFF 

Brickblock1:
Would be used instead of 0xFF with properties: draw_pylon_tiles, hide_wire_tiles and non_traversable_tiles.


PeterN:
I think you mean constant, not consistent.

Brickblock1:
> I think you mean constant, not consistent.

I did should be fixed now

glx22 (approved):

PR #285, opened 2023-04-18
Fix: [Actions] nmlc version could not be determined during regression

glx22:
Regression log is very noisy with each test triggering
```
fatal: No names found, cannot describe anything.
fatal: No names found, cannot describe anything.
Git checkout found but cannot determine its version. Error:  Command '['git', '-C', '/Users/runner/work/nml/nml', 'describe', '--tags', '--long']' returned non-zero exit status 128.
Git checkout found but cannot determine its version. Error:  Command '['git', '-C', '/Users/runner/work/nml/nml', 'describe', '--tags', '--long']' returned non-zero exit status 128.
```

Fetch more data during checkout so `git describe` can do its job.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -26,6 +26,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+
+    - name: Checkout tags
+      uses: openttd/actions/checkout@v2
+      with:
+        with-tags: true
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4


.github/workflows/regression.yml
```diff
@@ -26,6 +26,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
```
LordAro, line 30: Why is this required?

glx22: Because with default depth (1), you only get the merge commit, and its relation with last tag is impossible to find

rubidium42 (approved):

PR #286, opened 2023-04-19
Add: --no-32bpp and --no-extra-zoom option to skip some alternative sprites

glx22:
Following a discord conversation.
* zephyris
Hey, nmlc question for you all. Is there an argument which makes it ignore alternate zoom sprites? Similarly, make it ignore 32bpp alternates of sprites?
* zephyris
I'm hoping to make multiple versions of a grf at build time - file sizes get pretty enormous when including 32bpp and not everyone will need/want that.
* glx
I think it should be possible to add nozoom and no32bpp flags to nmlc

And it is indeed possible :)

### DIFF
--- a/nml/ast/alt_sprites.py
+++ b/nml/ast/alt_sprites.py
@@ -30,8 +30,10 @@ zoom_levels = {
     "ZOOM_LEVEL_OUT_4X": 4,
     "ZOOM_LEVEL_OUT_8X": 5,
 }
+allow_extra_zoom = True
 
 bit_depths = {"BIT_DEPTH_8BPP": 8, "BIT_DEPTH_32BPP": 32.0}
+allow_32bpp = True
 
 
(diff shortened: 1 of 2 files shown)


nml/ast/alt_sprites.py
```diff
@@ -30,8 +30,10 @@
     "ZOOM_LEVEL_OUT_4X": 4,
     "ZOOM_LEVEL_OUT_8X": 5,
 }
+allow_extra_zoom = True
 
 bit_depths = {"BIT_DEPTH_8BPP": 8, "BIT_DEPTH_32BPP": 32.0}
+allow_32pbb = True
```
rubidium42, line 36:
```suggestion
allow_32bpp = True
```
And everywhere where it was transposed I guess.

rubidium42 (approved):

PR #288, opened 2023-05-04
Change: Extend vehicle random bits to 16.

PeterN:
The top 8 bits will always be zero if the game does not support 16 bit vehicle. random data.

OpenTTD#10701 increased vehicle random bits from 8 to 16. This change does the NML side.

Not tested.

### DIFF
--- a/nml/actions/action2random.py
+++ b/nml/actions/action2random.py
@@ -62,13 +62,13 @@ class RandomAction2Choice:
 
 
 vehicle_random_types = {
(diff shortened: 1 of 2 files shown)


glx22 (commented):
I think https://github.com/OpenTTD/nml/blob/master/nml/actions/action2var_variables.py#L150 needs to be updated too.

glx22 (approved):

andythenorth:
nml spec updated https://newgrf-specs.tt-wiki.net/wiki/NML:Random_switch#Feature_and_Type

PR #293, opened 2023-05-28
Change: Support extended DCxx string range, and station names for IDs > 255

glx22:
Implements support for https://github.com/OpenTTD/OpenTTD/commit/f5394ed2ef32efdccb6312dd830d827e67d1311d and https://github.com/OpenTTD/OpenTTD/commit/bc7dfd7b46c0184dc4b2cc376556702ac0ac1019.

~~Draft because only half of it is done.~~

For stations I have 2 options, just switch to DCxx range, or use a mechanism to still use C5xx and C6xx ranges for IDs < 256 then fallback to DCxx.


michicc:
For the extended string range there also needs to be a way to reference them from a switch block.

As far as I can tell, a `string()` in a switch will always be allocated in the D0 range, even if it is used in the text stack and not as a CB result. Some kind of additional feature or automatic logic would be needed to also make it useable for e.g. the engine name callback.

glx22:
Yeah it allocates in D0 range and does some magic to return the unprefixed ID. Then authors have to readd the prefix in some cases (can't remember exactly where and why, but I'm sure I saw this being done)

glx22:
I used a very simple mechanism to switch range for station names depending on IDs.
For classnames I just decided to use only DCxx range, as in most cases many stations will use the same class.

nml/actions/action4.py
```diff
@@ -82,7 +82,11 @@ def skip_action9(self):
     0xD0: {"random_id": True, "total": 0x400, "ids": list(range(0xD3FF, 0xCFFF, -1))},
     # Misc. persistent text ids, used to set properties.
     # Use Ids DC00..DCFF first to keep compatibility with older versions of OTTD.
-    0xDC: {"random_id": True, "total": 0x800, "ids": list(range(0xDBFF, 0xD7FF, -1)) + list(range(0xDFFF, 0xDBFF, -1))},
+    0xDC: {
+        "random_id": True,
+        "total": 0x2800,
+        "ids": list(range(0xFFFF, 0xDFFF, -1)) + list(range(0xDBFF, 0xD7FF, -1)) + list(range(0xDFFF, 0xDBFF, -1)),
```
frosch123, line 88: This contradicts the comment from line 84. Now 0xFFFF is the first string ID to be used, raising the OpenTTD requirements.

glx22: Actually the values are poped from the list, so `DCFF..DC00` must be at the very end, and that's done with the last range.

frosch123 (approved):

PR #297, opened 2023-06-27
Change: Use a single replacenew-type 'SIGNALS' instead of 3 different constants.

### LINKED ISSUE #290: `replacenew` of `PRE_SIGNAL` incorrectly requires 48 sprites
The `PRE_SIGNAL` type for `replacenew` is documented as requiring 48 sprites, but does not indicate that all need to be replaced together. However, attempting to replace only a subset gives an error.

See also: https://newgrf-specs.tt-wiki.net/wiki/NML:Replace_new_sprites

`replacenew (PRE_SIGNAL) { [] }`

gives:

`nmlc ERROR: "ogfx2e_extra.nml", line 1: Invalid sprite count for sprite replacement type 'PRE_SIGNAL', expected 48, got 1`

Similar behaviour occurs with non-empty sprites. Tested using nml 0.7.2.

frosch123:
# Motivation
Fixes #290.

Currently there are 3 `replacenew`-types to replace signal graphics:
* `PRE_SIGNAL`
* `PRE_SIGNAL_SEMAPHORE`
* `PRE_SIGNAL_SEMAPHORE_PBS`

They all create the same code.

# Description
Add a new type `SIGNALS`, not to be confused with the existing `NEW_SIGNALS`.
The 3 old types will be deprecated and removed from the documentation.

### DIFF
--- a/nml/actions/action5.py
+++ b/nml/actions/action5.py
@@ -55,9 +55,10 @@ class Action5BlockType:
 
 
 action5_table = {
-    "PRE_SIGNAL": (0x04, 48, Action5BlockType.FIXED),
-    "PRE_SIGNAL_SEMAPHORE": (0x04, 112, Action5BlockType.FIXED),
-    "PRE_SIGNAL_SEMAPHORE_PBS": (0x04, 240, Action5BlockType.OFFSET),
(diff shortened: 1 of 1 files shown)


glx22 (approved):
Makes sense

PR #301, opened 2023-08-04
Fix: Register named parameters earlier

glx22:
Named parameters are registered during `pre_process` step, which can be too late.
Move the registration to `register_names` step.


frosch123 (approved):

glx22:
I'm not even sure it's possible to actually see the effect of this PR without more changes.
If I use https://github.com/OpenTTD/nml/pull/302/commits/5a5b4818a0e180b05dbbd0ef368aada88d992ba1 (and a named param for house id in 030) without this PR I get "unrecognized identifier", with the PR it becomes "Expected a compile-time integer constant"

PR #302, opened 2023-08-04
Add: 'const' keyword to allow users to define new constants

glx22:
It's possible to use named parameters as constants, but it doesn't always work and it's not very efficient in the resulting GRF when it does work.

Introduce the `const` keyword and `const <name> = <value>;` syntax, which will define the constant `<name>` to the value `<value>`.

Constants should be usable like any of the predefined constants so I delayed some validations for `item` and `disable_item` as they were happening too early.


glx22:
Found a perfect use case in 030_house.

regression/030_house.nml
```diff
@@ -50,8 +50,8 @@ spriteset(brewery_spriteset_building) {
     tmpl_building_sprite(150, 60, 91, -60, "brewery_snow.png")
 }
 
-default_palette = PALETTE_USE_DEFAULT;
-/* Generic sprite layout that is used for all tiles of the building. Parmaeters:
+const default_palette = PALETTE_USE_DEFAULT;
+/* Generic sprite layout that is used for all tiles of the building. Parmeters:
```
LordAro, line 54: Parameters

PeterN:
Have I understand this correctly? When not marked as const, the named parameter may be allocated a GRF parameter which is then dereferenced when used instead of simply storing the value where used.

glx22:
Exactly, named parameters are just GRF parameters, with all the implied limitations in usage.

PeterN (approved):

PR #304, opened 2023-09-20
Change: always assume non-track station ground tile

glx22:
For station ground sprite, OpenTTD always adds a railtype offset, unless a special trick is used to prevent it (like in example station before this PR).

With this PR, it is now always assumed the station ground sprite is a non-track one unless ground sprite is `GROUNDSPRITE_RAIL_X` or `GROUNDSPRITE_RAIL_Y`.

This hides the underlying GRF spec, and should simplify authors' life.

I think railtype offset might still be applied if ground sprite is from a spriteset, but it most likely depends on how the NML is written (`sprite: spriteset;` vs `sprite: spriteset(offset)`) and I didn't test, just guessing from the code.


nml/actions/action2layout.py
```diff
@@ -296,6 +296,14 @@ def _validate_sprite(self, name, value):
         else:
             self.sprite_from_action1 = False
             if isinstance(value, expression.ConstantNumeric):
+                # For stations always assume a non-track tile ground, unless explicit rail ground sprite.
+                if (
+                    self.feature == 0x04
+                    and self.type == Action2LayoutSpriteType.GROUND
+                    and value.value != global_constants.constant_numbers["GROUNDSPRITE_RAIL_X"]
+                    and value.value != global_constants.constant_numbers["GROUNDSPRITE_RAIL_Y"]
```
frosch123, line 304: To match `SplitGroundSpriteForOverlay` https://github.com/OpenTTD/OpenTTD/blob/7157e9666498d6a500b718a363dbcf16ed62b667/src/station_cmd.cpp#L3025 this should also include the snow variants.
Though NML seems to miss constants for those.

PR #305, opened 2023-09-26
Fix a9a1a3e: Don't use station properties 1C/1D for IDs 00-FF

### LINKED ISSUE #311: Incorrect error when not providing a `name` property for waypoints.
I don't think waypoints, coded as a special class of station, can have a name.

If you code a station using the `"WAYP"` class for `ITEM_STATIONS` and provide a `name` property, then OpenTTD reports a NewGRF error: `Unknown action0 property`.

If you do not provide a `name` property, then `nmlc` reports the error `nmlc ERROR: "ogfx2_stations.nml", line XXX: Property 'name' is not set. Item will be invalid.`, but makes a functional NewGRF that OpenTTD seems happy with.

Presumably just needs a special case for reporting the error? Or is this an OpenTTD bug?

glx22:
Properties 1C/1D are supported only by OpenTTD 14+.
Also using C5xx range in prop 1C is not supported by OpenTTD.

With this PR C4xx/C5xx are always used for IDs<256 without using properties 1C/1D, and for IDs>=256 the properties are used and the strings are in DCxx range.

Fixes #311


PeterN (approved):
Looks sane enough. 255 as an ID wasn't allowed pre-extended IDs, but that is not relevant here.

PR #306, opened 2023-09-29
Change: reduce register consumption for spritelayouts

glx22:
Spritelayouts can consume a huge amount of registers, especially with stations as all the spritelayouts are merged into a single action 0.

https://gist.github.com/glx22/19446d63c287f050719c54cb099c4f8f is a station using 40 spritelayouts and requiring 136 registers (I had to increase the nml reserved range to make it compile).
As can be seen in the "Station Layout@registers" varact2 most of the registers are set to the same values, that was not an issue for non station spritelayouts but it's a huge waste of registers for stations.

With this PR, a new register is used only for when it's really needed.
https://gist.github.com/glx22/e749d7f074cb5eb53b4228fd8340e48a is the same station as in the previous gist, but it requires 8 registers only


WenSimEHRP:
This PR is necessary for station sets with complexed station tiles that requires a large amount of registers. This PR should be merged into trunk.

https://github.com/openttd-china-set/china-set-stations and https://github.com/WenSimEHRP/jppluswins requires this PR to compile

PeterN:
People are using this patch in the wild, might need to be updated given the waiting statuses.

PeterN:
Can there be a summary comment of what the LayoutRegisters class is doing?

It's a bit black-magic to me, but I think it is working out if a value already exists in a register and then reusing it, instead of just creating a new register?

glx22:
Yes the idea is to map values to registers, so if a value is used multiple times it doesn't create a new register on each use.

PeterN (approved):

PeterN (approved):

PR #309, opened 2023-10-07
Change: Use extended Action1s

glx22:
Each spriteset gets its own unique ID.
Removes the "all spritesets in a single sprite group/layout must have the same size" limitation.

Not really tested beside checking NFO output.


andythenorth:
Works for me in test of 

- nml CHIPS with unequal sized spritesets
- pre-existing Iron Horse
- pre-existing FIRS

PeterN:
First question... what is an extended action 1?
Does this need a specific version of OpenTTD or is it an older feature that just isn't used?

glx22:
https://newgrf-specs.tt-wiki.net/wiki/Action1#Extended_format present since OpenTTD 1.2.
The most important part is https://newgrf-specs.tt-wiki.net/wiki/Action1#first-set which allows to reduce duplication and the need for same size `spriteset` in a `spritelayout`.

andythenorth:
Still works for me :P

glx22:
Now includes changes from @PeterN's pr309-redux branch.

PeterN:
I'm not sure why the examples are being changed -- presumably they don't do anything that requires this new format as they work before. Makes me think something is not quite right.

glx22:
Examples are changed because IDs are not reused anymore.
Before the change it was always act1 starting from 0 followed by act2 (act2 only using most recent act1), now all spritesets are accessible from anywhere once defined and they get their own ID.

PeterN (approved):

PR #310, opened 2023-10-10
Replaced deprecated setup.py with setup.cfg

bjornwarmedal:
Also bumped version to 0.7.5, with packaging being the only change.

The old setup.py method is deprecated and will be removed in future python3 versions. The rationale behind this is that the setup.py script runs arbitrary code at install time.

Our setup.py file used to dynamically determine the current version at install time by querying git. While this makes it more forgiving as a developer (setting a new tag is enough, no version update in any file is necessary) it's generally bad practice.

I suggest setting a github action that fills this role, if possible. Otherwise we must remember to manually update the "version" attribute in the setup.cfg file when releasing a new version.


matthijskooijman:
@bjornwarmedal, AFAICS, this PR removes the custom build_py command, which would generate the `parsetab.py` and `lextab.py` files. I suspect this means that these files will now be autogenerated on first use, which will not work if the user has no write permission for this (e.g. system-wide install).

Did you consider this removal and its (presumed - I did not test) effects?

PR #312, opened 2023-11-26
Fix cleanup leaving generated files behind

matthijskooijman:
This causes an issue with the Debian package, where running a build and then a clean should leave the source tree in the same state as it started in (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1045259).

I'm about to ship these patches as part of the 0.7.4 Debian package, but it would be better to fix these upstream.

I'm not 100% sure that hardcoding the filenames to delete in `setup.py` is ideal, but it is pragmatic and probably not worth the trouble delegating the deletion all the way into parser.py and tokens.py where the original filenames are defined.


matthijskooijman:
I've updated the PR, adding the first commit to prevent having to import `distutils` directly (which is deprecated and expected to be removed in an upcoming python version).

glx22 (approved):

matthijskooijman:
Hm, it seems these changes were merged in 0.7.5, but were not mentioned in docs/changelog.txt. Should I have added a changelog line in my PR? Or was this just overlooked in the release? Maybe good to add it retroactively for future reference?

PR #314, opened 2023-12-12
Change: Allow creating 32bpp-only NewGRFs

PeterN:
The NFO requirement that the first sprite is 8bpp and 'normal zoom', was never actually enforced by OpenTTD itself which has no problem loading 32bpp-only NewGRFs.

NML allows 32bpp and zoom levels but only on alternative_sprites.

This change supports the use of zoom level and bit depth parameters on the spriteset() instruction. The format is similar to alternative_sprites();

spriteset(name[, image_file])
spriteset(name, zoom_level, bit_depth[, image_file[, mask_file]])

A short regression test is included.

Rationale: some NewGRF authors are already creating 32bpp-only sets with other tools that already allow it, whilst authors using NML are creating 32bpp-only with fake transparent 8bpp sprites. Given that we can't (and shouldn't) force authors to create content they don't want to, it is better to just remove the requirement for 8bpp sprites. This is beneficial as it means that OpenTTD may be able to detect when 8bpp sprites are missing which it cannot do if dummy 8bpp sprites are present.


nml/ast/spriteblock.py
```diff
@@ -87,26 +87,62 @@ def __str__(self):
...
+        self.zoom_level = 0
+        self.bit_depth = 8
+        self.image_file = None
+
+        if len(param_list) >= 3:
+            if isinstance(param_list[1], expression.Identifier) and param_list[1].value in global_constants.zoom_levels:
+                self.zoom_level = global_constants.zoom_levels[param_list[1].value]
+            else:
+                raise generic.ScriptError(
+                    "value for alternative_sprites parameter 2 'zoom level' is not a valid zoom level", param_list[1].pos
```
glx22, line 106: It should say `Spriteset-block`

nml/ast/spriteblock.py
```diff
@@ -87,26 +87,62 @@ def __str__(self):
...
+                )
+            if isinstance(param_list[2], expression.Identifier) and param_list[2].value in global_constants.bit_depths:
+                self.bit_depth = global_constants.bit_depths[param_list[2].value]
+            else:
+                raise generic.ScriptError(
+                    "value for alternative_sprites parameter 3 'bit depth' is not a valid bit depthl", param_list[2].pos
```
glx22, line 112: It should say `Spriteset-block` and there's a trailing `l`

nml/ast/spriteblock.py
```diff
@@ -87,26 +87,62 @@ def __str__(self):
...
                 )
+
+        if self.bit_depth == 32:
+            global_constants.any_32bpp_sprites = global_constants.allow_32bpp
+
+        if len(param_list) >= 5:
+            self.mask_file = param_list[4].reduce()
+            if not isinstance(self.mask_file, expression.StringLiteral):
+                raise generic.ScriptError(
+                    "alternative_sprites-block parameter 5 'mask_file' must be a string literal", self.mask_file.pos
```
glx22, line 135: It should say `Spriteset-block`

glx22 (commented):
And some flake8/black problems

glx22 (approved):

andythenorth:
FIRS and Iron Horse appear to build fine with this PR (both 8bpp DOS-palette, 1x zoom)

PR #318, opened 2024-02-02
Add: Support town_production_effect and town_production_multiplier.

PeterN:
Provides support for town production effect and multiplier properties of OpenTTD#11947

```
        town_production_effect:     TOWNPRODUCTION_PASSENGERS;
        town_production_multiplier: 1.0;
```

```
        town_production_effect:     TOWNPRODUCTION_MAIL;
        town_production_multiplier: 0.75;
```


glx22 (approved):
Maybe we need confirmation it works from an NML user.

2TallTyler:
This PR adds the new properties to ITI: https://github.com/2TallTyler/improved_town_industries/pull/104

A question: Do I need to bump the minimum required OpenTTD version for the GRF (using a custom error), or will the properties be ignored in older versions of OpenTTD?

PeterN:
You'll need a version check. NewGRF is not self-descriptive so older versions don't know how to skip properties they don't know.

PR #321, opened 2024-02-15
Fix: Include information about relevant block in ScriptError message

### LINKED ISSUE #238: NMLC error from Produce block incorrectly refers to switch-block
## Version of NML
0.6.1

## Expected result

Attempting to produce a non-whole number quantity of cargo with the Produce block, as in the example below, produces an NMLC error stating that this isn't possible.

Example:
```
produce (produce_oil_refinery,
	[PASS: 1; OIL_: 1;],      // Consume
	[RFPR: 0.5; FERT: 0.5;],  // Produce
	1                         // Run callback again
	)
```

## Actual result

The error incorrectly refers to a switch-block:

`nmlc ERROR: "lumberjack_industries.nml", line 962: This expression is not supported in a switch-block`

## Steps to reproduce

1. Try to make NewGRFs in the early morning while not being fully awake :)
2. Attempt to produce a non-whole number quantity of cargo with the Produce block
3. Read NMLC error

Pixel-Tony:
Fixes #238, which refers to the error message in Action2 parser not containing `produce-block` as a relevant block.

This is my first commit to this repository, I hope someone will guide me in case anything is not correctly presented.

### DIFF
--- a/nml/expression/base_expression.py
+++ b/nml/expression/base_expression.py
@@ -92,7 +92,7 @@ class Expression:
         @return: True if this expression can be calculated by advanced varaction2.
         """
         if raise_error:
-            raise generic.ScriptError("This expression is not supported in a switch-block", self.pos)
+            raise generic.ScriptError("This expression is not supported in a switch-block or produce-block", self.pos)
         return False
 
     def supported_by_actionD(self, raise_error):


Pixel-Tony:
After already submitting this PR I noticed that mentioned error text is also present in some other Action2-related code. Since produce-block is based on this action, would it make sense to supplement those cases with mention of this block too?

glx22 (approved):

PR #322, opened 2024-02-24
Change: add --no-palette-validation optional arg

andythenorth:
To defend authors against mixing Windows / DOS palettes in the same grf, nmlc will open all sprite files from the filesystem and validate their palette. This is a separate step to sprite encoding.

For certain cases, this is of no benefit, and just increases compile times. For example, with Iron Horse:
* sprites are generated via a separate PIL compile step prior to nmlc
* the DOS palette is set during that sprite generation step as a necessary part of creating paletted image files
* therefore the subsequent nmlc check is redundant

The nmlc palette check adds 1s to 2s to every Iron Horse compile for me, which is 5% to 10% of total compile time.

I previously tried to solve this in #254, by inferring when the sprites might not need processed, but that route was not viable, as it depended too much on assuming that we don't want validation if we're outputting nfo.  

This route provides an explicit arg flag for authors to suppress validation.

~~The names of the flag and option probably are too long and not the best choice.~~

### DIFF
--- a/nml/main.py
+++ b/nml/main.py
@@ -89,6 +89,7 @@ def parse_cli(argv):
         debug_parser=False,
         allow_extra_zoom=True,
         allow_32bpp=True,
+        disable_palette_validation=False,
     )
     opt_parser.add_option("-d", "--debug", action="store_true", dest="debug", help="write the AST to stdout")
     opt_parser.add_option("-s", "--stack", action="store_true", dest="stack", help="Dump stack when an error occurs")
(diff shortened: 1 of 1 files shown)


nml/main.py
```diff
@@ -224,6 +225,7 @@ def parse_cli(argv):
         "--no-extra-zoom", action="store_false", dest="allow_extra_zoom", help="Skip extra zoom alternative sprites"
     )
     opt_parser.add_option("--no-32bpp", action="store_false", dest="allow_32bpp", help="Skip 32bpp alternative sprites")
+    opt_parser.add_option("--forcibly-skip-sprite-processing", action="store_true", dest="forcibly_skip_sprite_processing", help="Skip sprite validation processing")
```
glx22, line 228: That's a quite long option name.

andythenorth: Changed to `--no-palette-validation`

nml/main.py
```diff
@@ -89,6 +89,7 @@ def parse_cli(argv):
         debug_parser=False,
         allow_extra_zoom=True,
         allow_32bpp=True,
+        forcibly_skip_sprite_processing=False,
```
glx22, line 92:
```suggestion
        disable_palette_validation=False,
```

glx22 (approved):

PR #326, opened 2024-04-21
Change: Reduce D0xx usage where possible

glx22:
D0xx range is very small, and must be used for CB results.

Some CB allow use of {STRING} and text stack in the returned D0xx, and {STRING} supports DCxx in text stack. But filling the text stack will still allocate D0xx when a DCxx is desired.

Try to detect when we might be filling text stack and use DCxx range in this case.


PeterN:
Sounds reasonable. Do we need a unit-test for this?

PeterN (approved):

PR #330, opened 2024-05-18
Fix `nearby_tile_same_grf`, `nearby_tile_other_grf` and `nearby_tile_original_gfx`, they all compared to wrong constants

ahyangyi:
…original_gfx`, they all compared to wrong constants


ahyangyi:
Relevant grfspec page:

https://newgrf-specs.tt-wiki.net/wiki/VariationalAction2/Stations

ahyangyi:
Broken test is false alarm, should be fixed by #331...

LordAro:
I think it would be less confusing to reorder the definitions so that the numbers are still 0, 1, 2

ahyangyi:
I wanted to begin with a minimal delta, but the new order does make sense.

glx22 (approved):

PR #331, opened 2024-05-18
Update: Python versions for regression

glx22:
`macos-latest` doesn't have python 3.7 or 3.8 (though regression in PRs fails only for missing 3.7).
Test with more recent python version for all targets.


ahyangyi:
Yep, this causes all recent PRs to fail tests...

LordAro (approved):

PR #335, opened 2024-08-15
Add: Support for station properties 0E and 0F

glx22:
Since station tile types are no longer limited to the 8 original ones, it's time to implement support for property 0E/0F.

All the basics are validated I think, might need more work to prevent duplicated platform/length combo.

I also force the type to be even without notice.


regression/040_station.nml
```diff
@@ -123,6 +123,15 @@ item (FEAT_STATIONS, basic_station, 255) {
             bitmask(STAT_TILE_NOWIRE),
             bitmask(STAT_TILE_PYLON, STAT_TILE_BLOCKED),
         ];
+        station_layouts: [
+            [
+                [2],
+            ],
+            [
+                [4, 5],
+                [6, 7]
```
PeterN, line 132: Should this silently convert to 04 04 06 06 or throw an error?

glx22: For now it's silently converted, but an error or a warning is possible.

glx22:
Added warnings for non even tile types, they are still silently converted.
Also added check for layout redefinitions.

PeterN (approved):

PR #346, opened 2024-12-02
Update: changelog for 0.7.6

glx22:
I think I properly classified the commits.
I also checked release workflow still work (in my fork).


LordAro (approved):
I wonder if it's enough to make it 0.8.0, but eh.

PR #348, opened 2024-12-13
Add: --list-unused-strings to list the unused strings

glx22:
When you have a huge language file it's easy to end up with useless strings when modifying grf source.
The translators then have to translate useless strings.

Add an nmlc option to list unused strings.


andythenorth:
Works for me 😉

PeterN (approved):

PR #352, opened 2025-01-04
Fix: Properly check number of varaction2 ranges

glx22:
If a switch has more than 255 ranges, nmlc will trigger [this assert](https://github.com/OpenTTD/nml/blob/master/nml/output_base.py#L146).
Check earlier to report the error to the user in a more useful way.


andythenorth:
Works for me in testing.  Can't approve but works.

LordAro (approved):

PR #364, opened 2025-02-16
Fix 530e325f: 'gh' requires a git repository

glx22:
I completely forgot `gh` required a git repository when rewritting the release workflow, luckily it failed after uploading to pypi, so I could manually attach the source package to the github release.

I had different options to fix the issue:
 - Insert the following block before `gh` command
 ```
        # "gh" only works automatically if run from a git repository.
        git init
        git remote add origin ${{ github.event.repository.html_url }}
```
 - Move the `gh` command to `build_sdist` job, with the advantage of quicker availability of the source package on github release.
 
I chose to move the command.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,15 @@ jobs:
           name: cibw-sdist
           path: dist/*.tar.gz
 
-  upload:
+      - name: Upload to github
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload ${{ github.event.release.tag_name }} dist/*.tar.gz
+
+
+  publish:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/release.yml [hunk 1]
```diff
@@ -57,10 +65,8 @@ jobs:
       - name: Publish wheel and source distribution
```
frosch123, line 65: comment is out of date

glx22: What do you mean?

frosch123: It does not longer publish the source here.

glx22: It still does on pypi

frosch123: Oh, right...

frosch123 (approved):

PR #359, opened 2025-01-18
Add: Support for NewGRF badges.

PeterN:
Add support for NewGRF badge feature.

Use `badge_table { }` to create the translation table.
Use `badges` property to assign a list of badges to an entity.

See 042_badges.nml for slightly more info.

PR #378, opened 2025-05-30
Support vehicle var 0x65

andythenorth:
Supports [#14312.
](https://github.com/OpenTTD/OpenTTD/pull/14312)

Branched from #359 as depends on badge_parameter


PeterN (approved):
Probably

Rito13:
Both NFO and NML docs for var 0x65 now exist.

PR #385, opened 2025-09-01
Add: necessary properties in action0 in order to support bridges over station feature from OpenTTD 15.0-beta3

Rito13:
Allows to use bridges over stations feature for rail stations.
Do not contain necessary additions to allow same feature for road stops and other station types.
Remains compatible with OpenTTD 14.0 and older.

### DIFF
--- a/nml/actions/action0properties.py
+++ b/nml/actions/action0properties.py
@@ -16,6 +16,7 @@ with NML; if not, write to the Free Software Foundation, Inc.,
 import itertools
 
 from nml import generic, nmlop, global_constants
+from nml.global_constants import constant_numbers as global_constant_numbers
 from nml.expression import (
     AcceptCargo,
     Array,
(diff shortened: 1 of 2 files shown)


nml/actions/action0properties.py
```diff
@@ -818,6 +818,17 @@ def station_tile_flags(value):
...
 
+def array_for_station_properties(prop_num, list_type="Flag"):
+    def __call__(value):
+        if not isinstance(value, Array):
+            raise generic.ScriptError(list_type + " list must be an array", value.pos)
+        if len(value.values) % 2 != 0:
+            generic.print_warning(generic.Warning.GENERIC, list_type + " list does not have even length", 0)
+        return [VariableByteListProp(prop_num, [[flags.reduce_constant().value for flags in value.values]], True)]
+
+    return __call__
```
LordAro, line 829: don't use reserved names for non-reserved things ;) . `f` would be fine

nml/actions/action0properties.py
```diff
@@ -818,6 +818,17 @@ def station_tile_flags(value):
     ]
 
 
+def array_for_station_properties(prop_num, list_type="Flag"):
```
LordAro, line 821: not really any purpose in `list_type` having a default value

nml/actions/action0properties.py
```diff
@@ -818,6 +818,17 @@ def station_tile_flags(value):
...
+    def __call__(value):
+        if not isinstance(value, Array):
+            raise generic.ScriptError(list_type + " list must be an array", value.pos)
```
LordAro, line 824: generally speaking i'd say use f-strings instead of concatenation. But I suspect nml doesn't really use them because it's too old...

nml/actions/action0properties.py
```diff
@@ -818,6 +818,17 @@ def station_tile_flags(value):
...
+        if len(value.values) % 2 != 0:
+            generic.print_warning(generic.Warning.GENERIC, list_type + " list does not have even length", 0)
```
LordAro, line 826: the function above combines this with the array check. any particular reason not to do the same here?

Rito13: I imagine that it can be useful to only specify x direction and leave it to nml to figure out y direction flags.

nml/actions/action0properties.py
```diff
@@ -818,6 +819,36 @@ def station_tile_flags(value):
     ]
 
 
+def array_for_station_properties(prop_num: int, list_type: str, prefered_size: int, invert_function: callable):
+    def generated_function(value):
+        if isinstance(value, ConstantNumeric):
+            value = Array([value, invert_function(value)] * prefered_size, value.pos)
```
glx22, line 825: This is not a good idea to me.
All tiles having the same pillar placement limitation is probably quite rare.
And always defining 20 tile types feels wasteful.

nml/actions/action0properties.py
```diff
@@ -818,6 +819,36 @@ def station_tile_flags(value):
...
+        (global_constant_numbers["BRIDGE_PILLAR_CORNER_N"], global_constant_numbers["BRIDGE_PILLAR_CORNER_E"]),
+        (global_constant_numbers["BRIDGE_PILLAR_CORNER_E"], global_constant_numbers["BRIDGE_PILLAR_CORNER_N"]),
+        (global_constant_numbers["BRIDGE_PILLAR_EDGE_NE"], global_constant_numbers["BRIDGE_PILLAR_EDGE_NE"]),
+        (global_constant_numbers["BRIDGE_PILLAR_EDGE_SE"], global_constant_numbers["BRIDGE_PILLAR_EDGE_SW"]),
+        (global_constant_numbers["BRIDGE_PILLAR_EDGE_SW"], global_constant_numbers["BRIDGE_PILLAR_EDGE_SE"]),
+        (global_constant_numbers["BRIDGE_PILLAR_EDGE_NW"], global_constant_numbers["BRIDGE_PILLAR_EDGE_NW"]),
+    ]:
+        if value.value & (1 << i_flag):
+            out.value += 1 << o_flag
+    return out
```
PeterN, line 849: This logic should not be here, it's a bad idea to assume that the flags can be flipped.

nml/actions/action0properties.py
```diff
@@ -843,6 +874,10 @@ def station_tile_flags(value):
     "name":                  {"size": 2, "num": (256, -1, 0x1C), "string": (256, 0xC5, 0xDC), "required": True},
     "classname":             {"size": 2, "num": (256, -1, 0x1D), "string": (256, 0xC4, 0xDC)},
     "tile_flags":            {"custom_function": station_tile_flags},  # = prop 1E
+    "minimum_bridge_height": {"custom_function": array_for_station_properties(0x20, "Bridge heights", 10, lambda v: v)},
+    "bridge_pillars_flags":  {
+        "custom_function": array_for_station_properties(0x21, "Flag", 10, invert_bridge_pillars_flags)
+    },
```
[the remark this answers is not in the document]
Rito13: Is 0xFFFF better then?

nml/actions/action0properties.py
PeterN: Because NML shouldn't be assuming anything. There's no rule that says sprites should be "inverted".

Rito13: [Here](https://newgrf-specs.tt-wiki.net/wiki/NML:Stations#Rail_continuation) picture clearly says that other values are invertible.
Station_var45

PR #390, opened 2025-09-14
Add: Extend FEAT_TRAIN track_type for prop 34

glx22:
https://github.com/OpenTTD/OpenTTD/pull/14357 added prop 34.

Extended `track_type` syntax to also accept an array of railtype labels.
Prop 05 is used when there's only one label, prop 34 for more than one.


andythenorth:
Confirmed working as expected in my local test.

PR #419, opened 2026-03-12
Add: consist_current_power in varact2 for trains

WolfieFox:
OpenTTD has had these vars exposed since r3594 (Feb 2006) so I think we should be good to go on using them

### DIFF
--- a/nml/actions/action2var_variables.py
+++ b/nml/actions/action2var_variables.py
@@ -180,6 +180,7 @@ varact2vars_trains = {
     'current_railtype'    : {'var': 0x4A, 'start':  0, 'size':  8},
     'current_max_speed'   : {'var': 0x4C, 'start': 0, 'size': 16, 'value_function': value_mul_div(0x4786, 0x10000)},
     'vehicle_is_in_depot' : {'var': 0xE2, 'start': 7, 'size':  1},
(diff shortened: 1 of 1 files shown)


WolfieFox:
Valid, renamed the var to match.

2TallTyler:
Needs updating to squash the fixup commit, and rebasing so the Black CI check passes.

2TallTyler:
Commit name doesn't match the format (but the PR title is correct). 🙂

WolfieFox:
> Commit name doesn't match the format (but the PR title is correct). 🙂

Damnit, the rebase command prompted me three times and I didn't even parse it the second time. One moment...

2TallTyler (approved):
This variable is not published in NewGRF specs, but I found it. 🙂 

For others' reference: the source code exposes it in `vehicle specific properties` -> `VEH_TRAIN` -> `case 0x74` (offset by `0x80`):
https://github.com/OpenTTD/OpenTTD/blob/master/src/newgrf_engine.cpp#L924

If you're willing, this could use adding to the specs.
https://newgrf-specs.tt-wiki.net/wiki/VariationalAction2/Vehicles

If/when this is merged, it will also need to be added to the NML specs.
https://newgrf-specs.tt-wiki.net/wiki/NML:Vehicles

PR #431, opened 2026-05-28
Fix b86d930: incorrect sprite ID limit, and reuse old IDs when limit is reached

glx22:
#309 replaced a global limit of `0xFF` sprites per action1 block with a max id (`0x3FFF` or `0xFFFF` depending on feature) stored in each collection. And `nmlc` crashes once the limit is reached.

Actually all features should have `0x3FFF` as max id.
And crashing when limit is reached is not nice, it's better to start reusing old IDs.


andythenorth:
Works for me.  Resolves the failure cases that Iron Horse exposed.

PeterN (approved):

PR #433, opened 2026-06-08
Add: Missing features to random switch.

Rito13:
When road stops and badges were added the data for random action 2 was not.

Add TILE and SELF for road stops matching behaviour in stations. However road stops have 8 bits in TILE when stations have only 4 (based on https://newgrf-specs.tt-wiki.net/wiki/RandomAction2#Triggers)

Add empty dictionary for badges to generate pretty errors instead of internal ones.

### DIFF
--- a/nml/actions/action2random.py
+++ b/nml/actions/action2random.py
@@ -90,7 +90,7 @@ random_types = {
     0x0A: {"SELF": {"type": 0x80, "param": 0, "first_bit": 0, "num_bits": 16, "triggers": False}},
     0x0B: {},
     0x0C: {},
-    0x0D: {},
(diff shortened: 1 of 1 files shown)


glx22:
Not sure, but 0x0D being empty might be wrong.
I think it's should be `"SELF": {"type": 0x80, "param": 0, "first_bit": 0, "num_bits": 16, "triggers": False}` based on doc.

Rito13:
Based on [newgrf_airport.cpp L198](https://github.com/Rito13/OpenTTD/blob/master/src/newgrf_airport.cpp#L198) and [newgrf_spritegroup.cpp L268](https://github.com/Rito13/OpenTTD/blob/0149e4a3498606f402c6f0f943922ce6a805f187/src/newgrf_spritegroup.cpp#L268) I think that NFO specs are correct.

PR #434, opened 2026-06-22
Fix: Max number of spritesets would not error appropriately

Rito13:
In python list indexes are in range 0..(len(list) - 1).
Therefore custom_spritesets[len(custom_spritesets)] will fail.

Replace the invalid `>` with `>=`.


LordAro:
The "pythonic" way would be a try/except IndexError

LordAro (approved):

PR #438, opened 2026-08-15
Change: increase industry tile ID limit

andythenorth:
Industry tile ID limit increased from 254 to 511 in OpenTTD https://github.com/OpenTTD/OpenTTD/pull/12183

### DIFF
--- a/nml/actions/action0.py
+++ b/nml/actions/action0.py
@@ -199,7 +199,7 @@ used_ids = [
     BlockAllocation(0, 15, "Bridge", False),
     BlockAllocation(0, 4095, "House"),
     BlockAllocation(0, -1, "Global", False),
-    BlockAllocation(0, 254, "Industry Tile"),
+    BlockAllocation(0, 511, "Industry Tile"),
     BlockAllocation(0, 127, "Industry"),
     BlockAllocation(0, 63, "Cargo"),
     BlockAllocation(0, -1, "Sound"),


Rito13:
Why do we need to allow up to 511? 

Base on https://github.com/OpenTTD/OpenTTD/blob/df55704e7d336467df739c861d86ac09695fcba1/src/newgrf_commons.cpp#L104 I think that grfs are only allowed to define new industry tiles in range 175..511, any more tiles will get invalid id and will not be defined. That would be 337 free ids for newgrfs.

andythenorth:
> Why do we need to allow up to 511?
> 
> Base on https://github.com/OpenTTD/OpenTTD/blob/df55704e7d336467df739c861d86ac09695fcba1/src/newgrf_commons.cpp#L104 I think that grfs are only allowed to define new industry tiles in range 175..511, any more tiles will get invalid id and will not be defined. That would be 337 free ids for newgrfs.

Peter's comment here seems related https://github.com/OpenTTD/OpenTTD/pull/12183#issuecomment-1985890584

PeterN:
It's kinda arbitrary. The local ID used by the NewGRF doesn't directly relate to IndustryTile ID that will be used, so technically we could allow any IDs up to 65535 and as long as there are not too many defined it would still work.

But also you could use up 175 IDs to disable the original industry tiles, and then the remaining 337 to define your new ones.

andythenorth:
I'd be inclined to leave it at 511, I don't think nml should be trying to guard every case.

PeterN (approved):

PR #439, opened 2026-08-21
Add: Output warning if a similar name (case-insensitive) has been defined before

### LINKED ISSUE #427: Item IDs aren't checked for capitalisation
A little more context on discord https://discord.com/channels/142724111502802944/337701432230805505/1506201143842181260 but basically if I define the ID list as 

```
// Unit Wagons (Rail), available ID range: 13000-13001 (hex 0x32C8..0x32C9)
#define ID_RANGE_UNIT_WAGONS_RAIL 0x32C8..0x32C9
item(FEAT_TRAINS, item_mu_MU_Wagon_Unpowered, 13000) {}
item(FEAT_TRAINS, item_mu_MU_Wagon_Powered, 13001) {}
```

note non-lowercase above

but then define the individual item lowercase

```
item(FEAT_TRAINS, item_mu_mu_wagon_unpowered) {
    property {
        name: string(str_mu_mu_wagon_unpowered);
        
```

then the nml will compile without error but will cause silent failures in switches that depend on vehicle ids such as attachment rules.

PeterN:
In #427 an author had inadvertently mixed up cases of item names and experienced unexpected behaviour. In NML item names are case-sensitive, so using different cases results in a different item.

This PR adds a check for existing item names that differ only by case, and outputs a warning if found.

Item names are still case-sensitive, so there is no behaviour difference and therefore existing sets that rely on this should not break.

Closes #427.

> nmlc warning: "nml_output/043_name_case.nml", line 15: Item with name 'item_mu_mu_wagon_unpowered' similar to existing name 'item_mu_MU_Wagon_Unpowered'


LordAro (approved):
Strictly speaking, this isn't actually checking that the warning is output as expected. Dunno how much effort it would be to add support for such things (some sort of comment pragma, perhaps?)

PeterN:
> Strictly speaking, this isn't actually checking that the warning is output as expected. Dunno how much effort it would be to add support for such things (some sort of comment pragma, perhaps?)

Possible by capturing the stderr output and comparing it against expected stderr output, but that seems scope creep.

